PR #153, opened 2018-11-23
place tabs on top of windows, allows both viewers to be at same height

alcomposer:
The transport controls of both viewers are not aligned if the panel tabs are placed at the bottom. This allows both viewers transports to be aligned.


itsmattkc:
Already added in 4a7abcdeaab273caeeb1e619373d967d9c404527 but thanks for the PR!

PR #169, opened 2018-12-18
Timelineheader rightclick menu clear In/Out Points

alcomposer:


alcomposer:
Squashed and removed unused function

PR #182, opened 2018-12-21
Update README.md

alexmyczko:


alexmyczko:
The difference is if it's an official debian package, ubuntu gets a copy and users can just install it with apt-get or gnome-software or whatever. And it's built on more architectures than you provide (which is probably not much relevant anymore).

itsmattkc:
How do you mean "official Debian package"? Are you suggesting people wouldn't have to `add-apt-repository`?

alexmyczko:
Yes people would just run apt-get install olive,
first step done
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=917170
second one too
https://ftp-master.debian.org/new.html
next would be getting accepted, keep it updated with upstream,
and once in debian, 5 to 10 days, Ubuntu will have it too

itsmattkc:
Okay cool, thanks for setting this up. I'll be able to properly test it soon but it's more likely I'll put it on https://olivevideoeditor.org/linux.php than in the README.

itsmattkc:
Apologies for the lateness, link has been added to https://olivevideoeditor.org/linux.php

PR #195, opened 2018-12-24
antialias playhead

alcomposer:


itsmattkc:
Would it be possible to re-target this for the `master` branch? I've decided to make that the main development branch for now.

Issue #185, opened 2018-12-22
Add make install target


eszlari:
Having a `make install` target would help packaging under Linux.

itsmattkc:
Merged

PR #196, opened 2018-12-24
Add Linux metadata

eszlari:


itsmattkc:
I'm also considering changing the Qt target from `Olive` to `olive-editor`, which I know would be a pain in the ass but I feel like it's more sensible and it should probably be done sooner rather than later if so.

eszlari:
> I don't think I would use org.olivevideoeditor.Olive

Please make it spec compliant:
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

> I assume adding these addresses #185 ?

No, you would have to add somthing like
```
unix:!mac:target.path = $$PREFIX/bin
INSTALLS += target

effects.files = $$PWD/effects
unix:!mac:effects.path = $$PREFIX/share/olive-editor
INSTALLS += effects
```
See e.g.: https://github.com/mltframework/shotcut/blob/master/src/src.pro

eszlari:
Note, that this blocks submitting the flatpak to flathub: https://github.com/flathub (They don't support renaming yet).

Also note: This is not related to https://github.com/olive-editor/olive/issues/197 - the target / binary can be renamed at any time without affecting the flatpak.

itsmattkc:
What blocks submitting to flathub? (sorry I don't understand the context)

itsmattkc:
As in calling it something that isn't org.olivevideoeditor.Olive? I will concede it if it causes issues.

If you could add the full spectrum of icons (currently only in the `ubuntu` branch) and ensure this functions with the different binary name (which I'll probably do later today), I'll go ahead and merge.

PR #209, opened 2018-12-27
Build and upload AppImage, closes #159

### LINKED ISSUE #159: Add support for Flatpak packages
http://docs.flatpak.org/en/latest/introduction.html#reasons-to-use-flatpak
https://github.com/flathub/flathub/wiki/App-Submission

probonopd:
This PR, when merged, will compile this application on [Travis CI](https://travis-ci.org/) upon each `git push`, and upload an [AppImage](http://appimage.org/) to your GitHub Releases page.

Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions
- Can be listed in the [AppImageHub](https://appimage.github.io/apps) central directory of available AppImages
- Can double as a self-extracting compressed archive with the `--appimage-extract` parameter
- No repositories needed. Suitable/optimized for air-gapped (offline) machines

[Here is an overview](https://appimage.github.io/apps) of projects that are already distributing upstream-provided, official AppImages.

__PLEASE NOTE:__ For this to work, you need to enable Travis CI for your repository as [described here](https://travis-ci.org/getting_started) __prior to merging this__, if you haven't already done so. Also, You need to set up `GITHUB_TOKEN` in Travis CI for this to work; please see https://github.com/probonopd/uploadtool.

If you have questions, AppImage developers are on #AppImage on irc.freenode.net.

https://twitter.com/probonopd/status/1078110777649840128

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,35 @@
+language: cpp
+compiler: gcc
+sudo: require
+dist: trusty
+
+before_install:
+  - sudo add-apt-repository ppa:beineri/opt-qt593-trusty -y
+  - sudo sudo add-apt-repository ppa:jonathonf/ffmpeg-3 -y
+  - sudo apt-get update -qq
+
+install:
+  - sudo apt-get -y install qt59base qt59multimedia libavformat-dev libavcodec-dev libavfilter-dev libavutil-dev libswscale-dev libswresample-dev
+  - source /opt/qt*/bin/qt*-env.sh
+
+script:
+  - qmake CONFIG+=release PREFIX=/usr
+  - make -j$(nproc)
+  - make INSTALL_ROOT=appdir -j$(nproc) install ; find appdir/
+  - mkdir -p appdir/usr/bin/ ; cp olive-editor appdir/usr/bin/ # FIXME; "make install" should do this
+  - wget -c -nv "https://github.com/probonopd/linuxdeployqt/releases/download/continuous/linuxdeployqt-continuous-x86_64.AppImage"
+  - chmod a+x linuxdeployqt-continuous-x86_64.AppImage
+  - unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
+  - export VERSION=$(git rev-parse --short HEAD) # linuxdeployqt uses this for naming the file
+  - ./linuxdeployqt-continuous-x86_64.AppImage appdir/usr/share/applications/*.desktop -appimage
+
+after_success:
+  - find appdir -executable -type f -exec ldd {} \; | grep " => /usr" | cut -d " " -f 2-3 | sort | uniq
+  - # curl --upload-file Olive*.AppImage https://transfer.sh/Olive-git.$(git rev-parse --short HEAD)-x86_64.AppImage
+  - wget -c https://github.com/probonopd/uploadtool/raw/master/upload.sh
+  - bash upload.sh Olive*.AppImage*
+  
+branches:
+  except:
+    - # Do not build tags that we create when we upload to GitHub Releases
+    - /^(?i:continuous)/


probonopd:
An AppImage for testing is available at https://github.com/probonopd/olive/releases.

eszlari:
Disadvanteges of AppImages:

* not integrated in GNOME Software, KDE Discover or Mint Software Manager
* no sandbox included
* links against host libraries (can break)
* must be made executable (confusing for novices)
* needs more disc space than flatpak when installing multiple apps based on the same framework (e.g. Qt, GTK)
* harder to provide debug info than flatpak (which provides it as an easy installable extension)
* apps are harder to build (needs to be tested on ancient distributions)

probonopd:
While I respect your opinion, let me give some context for clarification. Disclaimer: I have been working on AppImage since over a decade.

> not integrated in GNOME Software, KDE Discover or Mint Software Manager

This can be seen as an advantage or disadvantage, depending on your view. GNOME Software, KDE Discover or Mint Software Manager contain software that is not coming directly from the original application author, so personally I prefer to get my software straight from the project's download page to ensure it is the latest, officially-built and officially-supported version.

> no sandbox included

If you want sandboxing, the you can use e.g., Firejail which has built-in support for AppImage.

> links against host libraries (can break)

Not if the host libraries maintain backward compatibility (which they should, or else we should consider them broken. This is like on any other operating system)

> must be made executable (confusing for novices)

Very easy:



Alternatively, you can install the optional `appimaged` daemon that does this automatically.

> needs more disc space than flatpak when installing multiple apps based on the same framework

Although it might be tempting to believe this, it turns out to be _wrong_ in many cases, mainly due to three reasons:

* Unlike Flatpak, AppImage does not ship KDE, Qt,... runtimes that contain _all of_ KDE, Qt. well-produced AppImages ship only the small subset of those frameworks that is _actually_ needed by that particular application. Makes a big difference!
* Flatpak and similar systems have the concept of runtimes, which are versioned. In practice, different applications may need different versions of the KDE, Qt,... runtimes. Needs a lot of space, especially since you have to download and store _all_ of it, not just the small subset an application actually uses
* AppImage keeps all data compressed on disk, hence reducing the space requirement further

So in practice, AppImage is often much more space efficient.

> harder to provide debug info than flatpak (which provides it as an easy installable extension)

The author who produces an AppImage can freely decide what to put inside. E.g., one could make a -debug version.

> apps are harder to build (needs to be tested on ancient distributions)

We recommend to build what goes into an AppImage on the oldest still-supported distribution, in order to maximize compatibility. This has nothing to do with the AppImage format, we would recommend the same if you would distribute your binaries in a zip file or on a CD-ROM

> extra tools (appimaged, AppImageUpdate) need to be installed for desktop integration & updates (distributions don't provide them in their repos)

These are all entirely optional. And we are eating our own dogfood, i.g. using AppImage to distribute them. Simply because distributions never carry the latest versions of those, in the way we want them to (e.g., due to distribution policies). Which I think strengthens the point why AppImage is great - we don't need the distributions to do anything.

probonopd:
Thanks for merging. However, the `GITHUB_TOKEN` environment variable si missing, please set it in the Travis CI settings of this project.

You can get one from https://github.com/settings/tokens. The permissions public_repo, repo:status, repo_deployment should be sufficient.

notramo:
@eszlari You wrote that it is a disadvantage of AppImage that it requires `appimaged` for desktop integration. Doesn't Flatpak and Snap require additional tools?

itsmattkc:
@probonopd Just added it, is everything working now?

itsmattkc:
Sorry I just noticed this setup doesn't copy the GLSL effects (xml, frag, and vert files from the effects folder). See a similar issue with a PR for `make install` https://github.com/olive-editor/olive/pull/210#issuecomment-450438291

I could possibly fix this myself if you'd prefer, the yml seems pretty straightforward.

probonopd:
It is not clear to me what I need to do, especially how is `<some dir>/effects` located relative to `<AppDir>/usr/bin/olive`.

itsmattkc:
With the `olive-editor` binary in `<AppDir>/usr/bin`, the files in the `effects` folder (not including the `internal` folder) would need to be copied to `<AppDir>/usr/bin/effects` (e.g. `<AppDir>/usr/bin/effects/boxblur.xml`). This possibly isn't ideal, and the current alternatives are placing them elsewhere and pointing Olive to the new directory with `OLIVE_EFFECTS_PATH`, or placing the `olive-editor` binary somewhere else and making a symbolic link in `<AppDir>/usr/bin/`.

probonopd:
Could you make the application search for them (also) in `(directory in which the main executable is)/../share/olive-editor/effects`?

PR #210, opened 2018-12-27
support make install for Linux

eszlari:


eszlari:
https://github.com/olive-editor/olive/pull/210

itsmattkc:
Hi, just got around to testing this and it works pretty well, however Olive doesn't find the GLSL effects in the location you've specified. At the moment, the only way to set the effects path is with environment variable `OLIVE_EFFECTS_PATH`

probonopd:
Is `$PREFIX/share/olive-editor/effects` the correct location?
Can Olive find `$PREFIX` at runtime relative to itself? E.g., `$PREFIX` is the directory above `bin/olive`. See http://doc.qt.io/qt-5/qcoreapplication.html#applicationDirPath

PR #212, opened 2018-12-27
Translation support for Appdata.xml

LeandroStanger:


prokoudine:
"NLE video editor" is a pleonasm, because it literally means "non-linear editor video editor". The same applies to translations.

Why not just spell it out: "non-linear video editor"?

PR #224, opened 2018-12-30
Move OS-specific files to /packaging

eszlari:


itsmattkc:
I'm concerned this might break the Debian/Ubuntu and Snap packages since (AFAIK) both packaging scripts check for those folders in the root directory as they are now.

eszlari:
In VLC's tree, snap isn't at the top level:
https://github.com/videolan/vlc/tree/master/extras/package/snap

eszlari:
FreeCAD also had debian in a subdir:
https://github.com/FreeCAD/FreeCAD/tree/releases/FreeCAD-0-17/package/debian

itsmattkc:
I've just tested and in their current form, moving them does break both of them. Presumably more modifications will need to be made before they can be moved out of the root folder.

eszlari:
I reverted the snap & debian commits. The other two commits should be non-controversial.

PR #273, opened 2019-01-08
allow next / previous keyframe buttons to jump to keyframe

alcomposer:


itsmattkc:
Using your example I was able to write a shared function that the timeline and keyframe view can both use (implemented in d3b02f3f53d7b34698705ca9a4b7d8e9724f1cc9). I don't need to merge, but this was very helpful nonetheless. Thanks!

PR #279, opened 2019-01-08
Build-files go into separate folders

jonno85uk:
I was looking into possibly assisting in this project and I wasn't keen on the filesystem-pollution when building the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,8 @@
 *.pro.user*
 Makefile
 .qmake.stash
+build/*
+snap/parts/*
+snap/prime/*
+snap/snap/*
+snap/stage/*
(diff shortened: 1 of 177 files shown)


jonno85uk:
The olive-editor binary is now located in a different sub-folder so integration tests are failing.

eszlari:
Why not simply build out of tree?

jonno85uk:
As in create a build-dir and then ../qmake && make ?

jonno85uk:
The pull request has gotten large so i'm closing.

I hope you can read this https://github.com/jonno85uk/olive/projects/2 as i'm going to continue on doing my thing whilst aligning with upstream because I like the program. However, I can't help but feel that if some of the accrued debt is tackled soon, this NLE will go the way of the rest.

PR #287, opened 2019-01-09
Project icons

pgilfernandez:
Hi,

My first commits with the project pane buttons with just icons, inclusion of first .svg files (thinking in future direct support for this files instead of using .png) and updated "add button" with better "dropdown corner".

I hope you like it
Cheers


itsmattkc:
I believe Qt can actually take SVGs directly, and that may be a good direction to go towards. However I'm uncertain if Qt correctly scales SVGs for HiDPI, and if it does, it probably differs depending on the platform/scaling technique.

itsmattkc:
Sorry about the late response, but yes I'm more than happy for you to redo the icons as you see fit. Thank you :)

pgilfernandez:
Hi,

I have almost finished creating all icons into an unified style and in SVG (I have exported too the actual .png) but I would like to know how do you want to proceed as you haven't merged this PR and I will create a new one with all the iconset.

Is something wrong or you want to do it anyhow on a different way?

itsmattkc:
Hi, yeah I haven't merged yet because I assumed there would be additions to this PR or a different PR later with all the icons. I don't mind how we proceed, but I felt all the icons in one PR would be simplest - let me know what works best for you.

pgilfernandez:
Ok, no problem.
Then I will create a new PR with everything inside.
What I don’t know is how to make Olive to load svg files, you will have to investigate it. In the meanwhile I will continue using the .png files

itsmattkc:
Yeah that's fine, once I merge new icons I can switch them over to SVGs

pgilfernandez:
Hi,
Well, it's done, how all the icons are ported to SVG and the Project pane has its own icons.
Try it out, I think it can be merged now.

I hope you like the redesigned icons.
=)

Issue #254, opened 2019-01-06
[Feature request] Themeable interface


pgilfernandez:
It would be great for users to have the option to choose the UI theme they prefer.
For that I would recommend:
- switch from png icons (raster) to svg icons (vector)
- don’t pack the icons inside the binaries but place them on /share/icons/ or similar folder so that they can be changed easily and “on the fly”. Well, this way the icons are loaded/imported at startup
- all colors, styles, backgrounds, etc moved to a .qss file (Qt permits it) which is also placed on a shared folder so that the user can edit it. Qss files are just text files with css language for styling Qwidgets. For reference FreeCAD does like that and works pretty fine (I’m the collaborator for their UI)

itsmattkc:
I'm interested in implementing this, my only concern is a lot of Olive's interface is set up with a QPalette and as far as I know there isn't an existing way to save/load those to a human-readable file (correct me if I'm wrong). Even so, QPalettes are not very complex so it wouldn't be difficult to write something up to save/load them.

pgilfernandez:
Well, I'm not an experienced programmer so what I'm going to tell you may be not purely right but a good direction to take:
When styling Qwidgets you have the option to use QPalette and style one by one them and the option of using stylesheets (stored as .qss files) which let you style all your Qwidgets at once. Of course you have to use rules (CSS rules) to reach the elements you want to style and not the rest.
As I have explained, Qt calls it "style sheets" ([Qt references](http://doc.qt.io/qt-5/stylesheet-reference.html)) and it's a very similar to web CSS styling. It has some limitation CSS hasn't but it is pretty powerful.
[Here](https://github.com/pgilfernandez/FreeCAD_stylesheets) you can find the work I did for [FreeCAD](http://freecadweb.org), now it's merged in master and the updates are done directly there.
For example, setting up all Qwidget text color and background would be like the following:
```
QWidget {
    color: white;
    background-color: red;
}
```
Very often is needed to style one specific Qwidget and not all existing ones, for that is very common to use IDs or Names for this Qwidget so that it can be styled with stylesheets like the following:
```
QWidget#MySpecialName {
    color: white;
    background-color: red;
}
```

I could help by creating the .qss styles but not how to making Olive read or swap them... but I guess you have the basic information now on how to do it... let me know when you want to try. You can even try with the ones I used for FreeCAD, they should work with any other Qt app.

itsmattkc:
[...]

PR #293, opened 2019-01-10
More proportional font size in the Action Search dialog

prokoudine:
This is a bit of arguing about tastes, but the action search dialog looks HUGE in comparison to the rest of UI :)

How about we decrease the multiplication factor down to 1.2? That makes the dialog's text slightly larger than the rest of UI text to make it stand out, but shows more content.


Symbian9:
This should be customized using "theme":

- https://github.com/olive-editor/olive/issues/254

And should NOT be hard-coded.

itsmattkc:
Heh I think I envisaged it like the Spotlight in newer macOS, but perhaps was a little overzealous. I'll merge this, but as @Symbian9 says it should probably be moved to a theme (or at least config) at some point.

PR #337, opened 2019-01-19
Fix memory leak from using libavcodec api

jonno85uk:
Identified 2 sources of memory leaks in io/previewgenerator.cpp

(cherry picked from commit 8c7eb8997660a70e1e58a4f89a82574931195e63)
Sequence class encapsulation work. Memory leak fixes work

(cherry picked from commit 7caaa778697ea2a84162f4ce523be8b6cef79647)

### DIFF
--- a/io/previewgenerator.cpp
+++ b/io/previewgenerator.cpp
@@ -1,4 +1,4 @@
-﻿#include "previewgenerator.h"
+#include "previewgenerator.h"
 
 #include "project/media.h"
 #include "project/footage.h"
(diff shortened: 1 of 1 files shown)


itsmattkc:
Hi, thanks for the PR! Unfortunately there were some flaws so I implemented fixes for these myself.

The `delete [] imgData` breaks thumbnails. QImage does not make a copy of that data when it's created, so deleting it corrupts it (see http://doc.qt.io/qt-5/qimage.html#QImage-3). You're correct however, that QImage does not clean up the data buffer itself in this scenario, which I hadn't realized before. According to documentation, it needs to be provided a function to do this which I've done in 30460c08a3b87f0ff205293267f37d8a710368c4.

Also `codec_ctx[]` is not a single AVCodecContext, it's an array of them. I think what you'd want to do is put `avcodec_free_context(&codec_ctx[i]);` underneath `avcodec_close(codec_ctx[i]);` in the for loop, which I've done in 6c92d65dbbc640699f6c8be2b183798f8cc39367.

Again thanks for pointing these out though, and feel free to point out any more like it.

Issue #356, opened 2019-01-22
[Feature Request] Extra keying effects


oc1024:
For now, Chroma key is the only keying effect available in Olive.

In addition to it, some more I'd like to see implemented:

- Luma key/using the clip itself as a mask is a common feature of other video editors, and would make many amateur special effects often used by YouTubers and other prosumers possible.
- Another keying effect I think would be interesting is a color threshold i.e. imposing a range for a certain component (HSL, RGB, YCbCr) and making other colors transparent. This would allow for custom bloom (by restricting luminance to the highest values and blurring the filtered image) and much more.

Some more I've thought of, but are guaranteed to become obsolete when masking is implemented:
- Filters for masks with simple shape (rectangles, ellipses, stars) - like Kdenlive's Alpha Shapes, maybe being superseded by something like Render > Shape when masking is done.
- Filter that takes an image mask, offset and scale as input - basically a limited version of full mask support.

goldenbait:
I would also suggest Difference matte - compare the pixels of two clips and remove everything that is static. Lets you isolate a moving character or object easily.

oc1024:
Implemented luminance key and selecting for color in #358 . For the simple shape masks, I don't have the skills necessary to implement them with anti-aliased rotation.

For image masks and difference mattes, I'd need documentation on how to write transitions (Kdenlive effects involving multiple images often hijack the transitions system like masks and blending modes). And they would be easily replaced by proper masking either way.

itsmattkc:
Much of this will be addressed in the node rewrite.

PR #358, opened 2019-01-24
Extra Keying Effects I: Luma Key and Color Selector

oc1024:
I took a shot at partially implementing what I requested at #356 . This enables bloom (Color Selector + Blur + HSL) and amateur fire VFX.

This is my first attempt at GLSL code; I reused a lot of code from other effects and there is likely room for optimization/style changes.

### DIFF
--- a/effects/colorsel.frag
+++ b/effects/colorsel.frag
@@ -10,6 +10,7 @@ varying vec2 vTexCoord;
 uniform float loc;
 uniform float hic;
 uniform int compo;
+uniform bool invert;
 
 float rgb2luma(vec3 c) {
     return (max(max(c.r,c.g), c.b) + min(min(c.r,c.g), c.b))/2.0;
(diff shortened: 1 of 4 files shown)


BrimsonBhin:
Works on `229d01a`. My only issue is that the min/max values are very sensitive. I would recommend scaling the thresholds from _**0-1**_ to  _**0-100**_, like the Chroma Key shader.

Nice work by the way

oc1024:
@BrimsonBhin Implemented 0-100 scaling for this effect. As you said, makes it consistent with the Chroma Key effect and it is now possible to input values other than 0% and 100% with the mouse.

Also added an invert checkbox. Not very useful for most applications, but piping it between invert effects wasn't efficient.

oc1024:
Olive's new blending mode infrastructure wants us to multiply RGB by A or else it'll overlay the colors instead. This commit will make the plugins compliant with this requirement, plus will allow the Color Finder shader to run on OpenGL 2.0.

itsmattkc:
@oc1024 I was about to mention, this was broken by the new blending algorithm. Glad you were able to discover and fix that. I'll test it properly tomorrow and should merge shortly after.

PR #368, opened 2019-01-25
Limit bezier handles

alcomposer:
Stop control point handles from being able to cross centre of control point, and flip over to the wrong side. This is the behaviour of handles in other software that uses bezier / graph editors.


itsmattkc:
Sorry about the late response, thank you for the PR and letting me know about this - I've used your code example to write a slightly different algorithm in 09033e9cdf11cb9edbaceec2b53af546e47191df

PR #370, opened 2019-01-25
"for" loop counters should not have floating type

jonno85uk:
When using a floating-point for loop counter, an accumulation of rounding errors may result in a mismatch between the expected and actual number of iterations.

Even if floating-point loop counters appears to behave correctly on one implementation, it may give a different number of iterations on another implementation.


itsmattkc:
Thank you for making the PR and informing me about this, I've chosen to remove the code entirely instead as it was messy and never actually used (it was a relic of a much earlier commit).

jonno85uk:
Yeah. I was trying to see how it got used in the code by putting a breakpoint there but it never hit. Wasn't sure if the code was supposed to be used.

PR #379, opened 2019-01-30
added initial german translation

naj59:
Most of the things are translated. I am struggling at some points so I hope that there is someone later who is going to help me out. Hope this is ok for the moment.


itsmattkc:
Looks good! I'll write a drop-down list to switch to it soon too.

naj59:
@itsmattkc no problem :-) you can also wait until more translations are added.

PR #408, opened 2019-02-02
Add initial Russian translation

prokoudine:
Most user-visible things (that are translatable) are now translated.


itsmattkc:
Awesome, thank you!

Issue #451, opened 2019-02-07
Olive icon is incorrect under KDE Plasma Wayland session


glaubersm:
Hi

I have Olive installed on Arch Linux running KDE Plasma desktop environment under Wayland.
As we can see in my screenshot, Olive shows a Wayland generic icon instead the correct icon.
Incorrect icon also appears when I press alt+tab to switch between apps.

![olive](https://user-images.githubusercontent.com/15967631/52423048-72e18780-2ade-11e9-8c35-49b8d37d874c.png)

itsmattkc:
I'm not very familiar with Arch/KDE. Which package are you using?

glaubersm:
I have both flatpak and AUR packages installed on my system, both are affected.

https://flathub.org/apps/details/org.olivevideoeditor.Olive

https://aur.archlinux.org/packages/olive/

elsandosgrande:
So, the icon in the windows is not registered in my installation of KDE Wayland either (Antergos), but the icon is registered by Latte Dock (the same can't be said for some other programs). 
Here are my [installed packages](https://github.com/olive-editor/olive/files/2842804/installed.packages.txt) if they may be of any use.

itsmattkc:
Does this issue also occur with the current AppImage?

glaubersm:
Icon is correct with AppImage.

itsmattkc:
My only guess is the icons are set up correctly for GNOME (and similar) but not for KDE.

@eszlari, do you know what might be causing this?

eszlari:
This should work:
```
QCoreApplication::setOrganizationDomain("olivevideoeditor.org");
QCoreApplication::setApplicationName("Olive");
```
(must use the same naming as the file `org.olivevideoeditor.Olive.desktop`)

itsmattkc:
Done in 33c53a2c7925d87fd1e51c24180ed234054be445, let me know if that solves this

eszlari:
You set `setOrganizationName` instead of `setOrganizationDomain`!

itsmattkc:
Whoops, fixed in 7f78f544f0dc0b4b8e4102d936e7670d7652eb49

glaubersm:
I have just updated the flatpak build and I can confirm that Olive icon is correct now.
Thank you all.

PR #481, opened 2019-02-12
linux/wayland: fix icon

eszlari:
https://github.com/olive-editor/olive/issues/451


eszlari:
Taken from https://github.com/flatpak/qt-flatpak-demo/blob/master/main.cpp

PR #486, opened 2019-02-12
Updated Russian translation

prokoudine:


prokoudine:
@itsmattkc 

I don't understand why failing to fetch some non-translation data fails the translation check :-/

itsmattkc:
@prokoudine I assume it's an unrelated, especially since it only affects one of the builds.

PR #509, opened 2019-02-14
Translations

elsandosgrande:
I have just one question. Why audio and video effects in all caps?

### DIFF
--- a/ts/olive_ar.ts
+++ b/ts/olive_ar.ts
@@ -803,681 +803,681 @@
 <context>
     <name>MainWindow</name>
     <message>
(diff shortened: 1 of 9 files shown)


itsmattkc:
> Why audio and video effects in all caps?

Mostly cosmetics. I thought it looked better.

elsandosgrande:
Umm, yeah, that doesn't look to well without an adequate typeface. I presume a fixed one is not given, because it does not look appealing in the font that it utilizes on my system by default. Perhaps change the case until you add fonts?

PR #513, opened 2019-02-14
Translations

elsandosgrande:
I'd suggest putting the ExportThread items in sentence case.
Also, what are `could not video encoder for %1` and `could not audio encoder for %1` supposed to mean exactly?

### DIFF
--- a/ts/olive_bs.ts
+++ b/ts/olive_bs.ts
@@ -321,93 +321,93 @@
     <message>
         <location filename="../dialogs/exportdialog.cpp" line="61"/>
         <source>Export &quot;%1&quot;</source>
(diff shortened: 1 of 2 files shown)


itsmattkc:
Olive relies on FFmpeg for encoding, which in turns relies on several other encoding libraries to function (e.g. libx264, theora, etc.) These encoders may or may not be installed on the system/compiled into FFmpeg. `could not X encoder` means that Olive/FFmpeg are aware of the codec but there is no library/software available to it to encode into that codec.

elsandosgrande:
@itsmattkc So, the software is basically not available. In that case, I'll correct my translations. Thank you for the information!

PR #519, opened 2019-02-16
File-dir restructure for unit-test example

jonno85uk:
This isn't a PR that I expect as being accepted, more of a demonstration of how it'd work.

Everything but main.cpp becomes a static library which is linked to when main.cpp of olive or unit-tests is built.

There's a QMAKE setting in the .pro file I can't remember which makes the library .pro a dependency of the olive/unit-test .pro
With the setting if a change is made in the library, the other projects ensure to build the library 1st. ATM, it doesn't do that so a rebuild of the olive/unit-test projects is needed to pick up the new library changes.

Hope that makes sense.


itsmattkc:
Sorry for the lateness. I can see it's quite a paradigm shift and I'll admit isn't something I'm familiar with, though I can follow it for the most part. It seems to be mostly changes to the .pro files to produce, as you say, the app as a library that will then be used to create the main app or a testing app. It's very interesting and could probably be done.

The examples here seem quite basic but I'm guessing the idea is you could test almost anything that takes input and produces output? I'm also guessing this is primarily useful for ensuring changes to one class won't affect the others that use it?

I think I'll have to read further about unit testing but this is certainly a helpful example. As you've said before, Olive could do with more distinct classes with a defined purpose and less reliance on other classes (which has become even more apparent as I refactor and document things) which I feel would make unit testing even more purposeful.

jonno85uk:
No worries. It it quite a change which is why I submitted it as an example. There is a simpler way but very cludgy, just copy-pasta the src+headers into the test .pro file.

Yes, i'm not sure I had any test-examples in this PR but for the ones I have in the master-branch, it was to help my understanding whilst refactoring code and to get the ball rolling.
Ideally you keep each test-case as small as possible and the tested-function even smaller still. Obviously, not everything can be made into black-box like functions so why some of the test classes are private friends of the class under tests. 
Sometimes its helpful to overload certain class methods by creating a subclass and testing that instead. For example Someclass::readFromDB could be overloaded to something trivial in order to satisfy some flow condition.

Unit-testing restrospectively can be a laborious, soul-sucking experience so I would only suggest, if you do intend on doing it, to use them to capture some identifiable small functionality whilst refactoring. Not test everything, because it will suck your will to live.

itsmattkc:
I see. So your recommendation is mostly to use them for refactoring to ensure classes/functions continue to produce correct output after modification? It's interesting and I can certainly see how it'd be useful, perhaps even more so as the classes become more modular. I appreciate the example so demonstrate how it it'd work. There may be a time when implementing these may be useful. Out of curiosity, are there use cases for unit testing assisting in increasing project scale or is it mostly useful for rewrites/refactors?

jonno85uk:
> I can certainly see how it'd be useful, perhaps even more so as the classes become more modular

Yes, currently it's difficult to test the code base due to the various dependencies and program state.

> Out of curiosity, are there use cases for unit testing assisting in increasing project scale or is it mostly useful for rewrites/refactors?

It's useful for both. It can help detect  regressions i.e. a unit-test capturing a failing state picked up in a bug report/issue.

PR #521, opened 2019-02-16
Translations

elsandosgrande:
Umm, what is the Graph Editor for?


jonno85uk:
> 
> 
> Umm, what is the Graph Editor for?

I've never used it but from what I can tell, it's for animation.

https://www.schoolofmotion.com/tutorials/intro-to-the-graph-editor-in-after-effects/

alcomposer:
Graph Editor = Bézier curve editor. It allows the user change the easing of control points with bézier curves.

jonno85uk:
> 
> 
> Graph Editor = Bézier curve editor. It allows the user change the easing of control points with bézier curves.

Is that the control points (keyframes, i presume) of effects?

alcomposer:
@jonno85uk yes. It allows the user to change the easing between keyframes. (it also allows the user to choose between using Bezier or Linear interpolation). Hence it is for Effects, (which is a loose term- as some effects can be generators) in the time domain.

elsandosgrande:
Oooh, I have never used a graph in that way, so I didn't get it at all. The only graph-looking keyframing in KDEnlive was the graph being on the actual video clip in the timeline, which isn't a bad system, at least compared to the current graph editor's touchpad controls (I might as well be rolling dice when it comes to manipulating that thing with zoom and all). Considering that I will close this pull request soon as will add new translations soon (homework is a pain), I just want to ask one thing before that. Should we move this to another feature request/discussion tracker?

Side-note
I was hoping an explanation would help me with the translations of the right-click menu items, but I'm still just as dumbfounded as to how to translate "Zoom to Show All" (is it zooming to "show all" with "s.a." being the object, or is it some other weird compound).

itsmattkc:
Hi, thanks for the update! Can you restrict your PR to just the files you've updated? (e.g. `ts/olive_bs.ts` and `ts/olive_sr.ts`)

Also to help clarify further, as others have said, the Graph Editor shows keyframes and allows you to adjust the timing/velocity of them. "Zoom to Show All" will set the zoom to show all current keyframes within the viewport. Think of it like a zoom to "fit". Similarly, "Zoom to Selection" will set the zoom to show all the keyframes currently selected. Let me know if you have trouble with any others.

elsandosgrande:
I am not sure how to restrict the times, as I pull from you (upstream) and that ends up queuing other times as well (I only pull in the case that there have been other commits before I was ready for pull request). I do it to minimize any chance of there being issues with the pull requests. 

Also, thanks for the clarification! I am currently at school (lunch break), but I'll get back to it tomorrow or today if possible (technical drawing is not my strong suit, but I'm slow in general, do it'll take some more time).

itsmattkc:
I assume you're using `git add .` and then `git commit`? Instead of `git add .` you should be able to add the files individually (e.g. `git add ts/olive_bs.ts`) before you commit so you don't end up committing all the other language files too. I'll just add them individually for this one but if you could limit future PRs to just your files that'd be great!

elsandosgrande:
I was using `git commit -a` (lazy bones here). Roger that Sir!

PR #527, opened 2019-02-18
Mark user-visible message for translation and update Russian translation to match latest changes

prokoudine:


itsmattkc:
Good catch, thank you!

PR #530, opened 2019-02-19
Simplification of debug.cpp:debug_message_handler

jonno85uk:
Something picked up by sonarcloud


itsmattkc:
Looks much better, thanks!

Issue #534, opened 2019-02-20
Crash on opening file in Media Viewer


mdmayfield:
*If you are requesting a feature, you can remove the following content. If not, it would help get a faster feedback on your issue.*

__Olive version:__  e0b2cf3b7af5b481f264b83111df8ffd7aceff65
__Source:__ git
__Operating system:__ Ubuntu MATE 18.10
__CPU:__ Intel Core i5-6300HQ
__GPU:__ Intel i915

### Detailed Description

On double-clicking a media file to view it in the Media Viewer, application crashes. See MKV in attached zip file for video demonstration.

### Steps to reproduce

1.  Open project with video media
2.  Double click on entry in the Project pane
3.  Application crashes (segfault)
[Double-Click-Media-Crash-20190220.zip](https://github.com/olive-editor/olive/files/2884795/Double-Click-Media-Crash-20190220.zip)

### Output log

Please see attached for more details.
```
2019-02-20T07:34:55 [ERROR] Failed to retrieve frame from buffersink. -11 (playback/cacher.cpp:547, void cache_video_worker(ClipPtr, long int))

Thread 1 "olive-editor" received signal SIGSEGV, Segmentation fault.
0x00005555555f09c0 in Media::get_type (this=0x6d8ce1e63ee5a000)
    at project/media.cpp:210
210		return type;
(gdb) bt
#0  0x00005555555f09c0 in Media::get_type() (this=0x6d8ce1e63ee5a000)
    at project/media.cpp:210
#1  0x0000555555672250 in SourcesCommon::mouseDoubleClickEvent(QList<QModelIndex> const&) (this=0x555555de0b10, selected_items=...)
    at project/sourcescommon.cpp:253
#2  0x00005555555dab10 in SourceTable::mouseDoubleClickEvent(QMouseEvent*)
    (this=0x555555dd0b00) at ui/sourcetable.cpp:76
#3  0x00007ffff5c6730f in QWidget::event(QEvent*) ()
    at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#4  0x00007ffff5d09e7e in QFrame::event(QEvent*) ()
    at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#5  0x00007ffff5e7b19b in QAbstractItemView::viewportEvent(QEvent*) ()
    at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
[...]

PR #536, opened 2019-02-20
Fix crash on double-click media item #534

mdmayfield:
Fix crash on double-click media item #534


itsmattkc:
This is definitely what that line of code is supposed to be, thanks for picking up on this

Issue #549, opened 2019-02-25
[Feature Request] Ctrl-wheel for zoom instead of Shift-wheel


mdmayfield:
It's more common (in my experience with Web browsers, image viewers, file managers, and other software) for Ctrl-mousewheel to be used for zoom, and Shift-mousewheel to be used to swap scrolling axes (horizontal/vertical).

For consistency with other applications, please consider making Ctrl the zoom modifier and Shift the axis modifier.

elsandosgrande:
One, I agree completely on the ctrl note. Two, it uses SHIFT to zoom? Three, you use SHIFT with a mouse wheel to scroll horizontally?

mdmayfield:
This is a complex situation that will need some judgment calls. There are several competing considerations:

- Some users have a mouse with only one wheel. Others have a touchpad with multi-axis scrolling.
- The most strongly expected behavior from a scroll wheel is vertical scroll.
- Most applications and graphical toolkits, across OSes (to my knowledge) scroll horizontally on Shift-scroll wheel.
- Some users know about Shift-scrollwheel to scroll horizontally; many don't.
- Many, but probably not a majority of, applications that have a primarily horizontal workspace (i.e. timeline-based editors) use the unmodified scroll wheel to scroll *horizontally* instead of vertically as "generally" expected.
- Olive currently does this (vert scrollwheel --> horiz scroll)

If it were up to me alone, here is what I would choose for scrollwheel behavior:

- Timeline: unmodified scroll = vertical scroll. Shift-scroll = swap axes. Ctrl-scroll = horizontal zoom.
  - Prioritizes "usual behavior" and is convenient for touchpad users
  - Slight disadvantage to mouse users
  - To mitigate that, tooltips or similar to raise awareness of Shift-scrollwheel

- Graph Editor: exactly the same behavior as the Timeline, same modifiers

When swapping axes, I would make the wheel motion that normally scrolls a webpage _down_ scroll the timeline to the _right_. (Same motion = "later" in the document/project, whether vertical or horizontal)

elsandosgrande:
@mdmayfield @itsmattkc still hasn't gotten the touchpad scrolling to work properly (#505), so it's there is no difference on that note at this time. Oh, and the vertical-gives-horizontal scrolling thing is also mentioned in the issue. 

[...]

PR #560, opened 2019-02-26
Rework timeline scrolling

mdmayfield:
This is an attempt to satisfy both mouse and touchpad use-cases, and multiple modifiers for axis swap:

- Mouse users usually want horizontal scrolling with the mouse wheel
- Touchpad users will want two-finger scrolling to move as expected, both horizontally and vertically
- Users accustomed to Qt will expect to use Alt/Option for axis swap
- Users accustomed to browsers and other apps expect Shift for axis swap
- Refactored the code a bit to be easier to read

With this patch, the mouse wheel will scroll horizontally if the pointer is within a section (audio vs video) of the timeline that doesn't have space to scroll vertically, or if that section is scrolled as far as it will go vertically in the direction the user is trying to scroll.

If there's room to scroll up or down, a mouse wheel will scroll vertically until it can't go further, after which it will scroll horizontally in the appropriate direction (up = left and down = right).

Two-finger scrolling on a touchpad works entirely as expected, with the exception that vertical two-finger scrolling behaves the same as mouse wheel scrolling (horizontal at ends of travel) as described above.

Either Shift or Alt can be used for axis swap.

As far as I could find, there is no straightforward way to tell the difference between scroll events sent from a physical mouse wheel vs. two-finger scrolling on a touchpad. If there is such a way, I'd recommend an approach using that instead of the behavior here.

Related to discussion in #549

### DIFF
--- a/ui/timelinewidget.cpp
+++ b/ui/timelinewidget.cpp
@@ -360,49 +360,81 @@ void TimelineWidget::dragMoveEvent(QDragMoveEvent *event) {
 }
 
 void TimelineWidget::wheelEvent(QWheelEvent *event) {
-  // ctrl used to toggle zooming instead of scrolling
+
+  // TODO: implement pixel scrolling
+
+  // "Scroll Zooms" false + Control up  : not zooming
+  // "Scroll Zooms" false + Control down:     zooming
+  // "Scroll Zooms" true  + Control up  :     zooming
+  // "Scroll Zooms" true  + Control down: not zooming
+
   bool ctrl = (event->modifiers() & Qt::ControlModifier);
+  bool zooming = (olive::CurrentConfig.scroll_zooms != ctrl);
 
-  //
-  // NOTE/FIXME: CURRENTLY disabling pixel scrolling because it needs more testing
-  //
+  // Qt automatically swaps axes behind-the-scenes on Alt. Allow Shift too.
 
-  /*if (!event->pixelDelta().isNull()) {
-        // if we got pixel scrolling data, prefer it over the angleDelta data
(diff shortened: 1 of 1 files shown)


mdmayfield:
Updated this with a couple improvements:

1) Zoom was way too fast on touchpads because of ignoring the "120 distance = 1 click" convention in Qt.
2) Zoom in/out was asymmetrical (1.5 vs. 0.75); this should be 1.33 and 0.75, or 1.5 and 0.67. It seemed to feel nicer at 1.33 and 0.75, so used that as a base.
3) Took advantage of both Shift and Alt together [double-reversing the polarity](https://youtu.be/pIpl3VxttmM?t=18) to implement an "only scroll vertically, no horizontal shenanigans" option

elsandosgrande:
I think the "scroll one way until impossible, then switch axis" is not very good, at least on a touchpad. I would suggest that behavior to be optional, while the no-auto-switch to be default.

mdmayfield:
@elsandosgrande If there were a way to tell if the user had a touchpad or mouse, I would agree. But then there's the case of a touchpad *without* horizontal scrolling.

It's a very delicate task to try to satisfy all use cases here; there's a lot of different configurations possible.

One possibility might be a checkbox that would make the default behavior the same as if both Alt and Shift are held - vertical scrolls only scroll vertically, and horizontal scrolls only scroll horizontally, no matter what. If a mouse user had that checked they would be required to hold Shift or Alt to scroll horizontally (which might be OK with a lot of people).

mdmayfield:
Gonna close this one for now - @elsandosgrande brings up some good points and I want to make sure that this is consistent with what I'm working on for the Graph Editor. Will open a new PR later.

Issue #505, opened 2019-02-13
Two-finger scrolling in the timeline


elsandosgrande:
I know this is a minor issue, but I can't scroll when hovering over the timeline. 

Basically, I can use horizontal two-finger scrolling in KDEnlive to easily scroll through the timeline without breaking my flow, but I always have to drop down to the scrollbar before I can use two-finger scrolling, but I might as well drag the bar at that point.

Post Scriptum
It is at this point in typing that I checked and yes, there is scrolling, but it's only vertical (like a mouse-wheel), so my issue stands.

itsmattkc:
You should be able to scroll horizontally too by holding Alt

elsandosgrande:
I'll check, but that is not necessary with KDEnlive (again, I did state that it was probably a minor issue, of course when considering the stage of development and all of the other issues in the backlog).

Post Scriptum
Nope, alt makes it not scroll, even horizontally.

itsmattkc:
Actually disregard my previous message.

Vertical scrolling on the timeline is supposed to scroll horizontally by default. You're saying this doesn't happen? What desktop environment are you using?

elsandosgrande:
I'm not talking about the axis of the timeline, but of my fingers. When I move my two fingers from-left-to-right, it doesn't scroll, but when I move my fingers up-down, it does scroll. 

Antergos KDE by the way.

oc1024:
What at @elsandosgrande wants, I think, is to be able to scroll the timeline horizontally with horizontal two finger scrolling instead of having to make the rather unintuitive association " fingers up and down = timeline left right".

elsandosgrande:
@oc1024 You're my man! It's so much easier in KDEnlive. Also, the scrolling feels like the sensitivity was turned down to a very low preset until I get to the scrollbar (I guess KDE itself takes over then).

mdmayfield:
I kind of expected that too, that two-finger horizontal scrolling would work. In some media applications, vertical scrolling zooms in/out and horizontal scrolling scrolls; personally, my preference would be:

#### "Natural" scrolling
- Two fingers move up on the touchpad: timeline scrolls to show later events
- Two fingers move down on the touchpad: timeline scrolls to show earlier events
- Two fingers move left on the touchpad: timeline scrolls to show later events
- Two fingers move right on the touchpad: timeline scrolls to show earlier events

#### Traditional scrolling
- Two fingers move up on the touchpad: timeline scrolls to show earlier events
[...]

Issue #550, opened 2019-02-25
[FR] Graph Editor navigation is inconsistent with other apps / other parts of Olive


mdmayfield:
Currently, in Olive build f4668e317a4fb39968a6bf3b03cc3d968b393c07 I am seeing this behavior in the Graph Editor:

- No scroll bars, no autoscrolling; no ability to scroll apparent in the UI
- No zoom buttons; no ability to zoom apparent in the UI
- Mouse wheel behaves very differently from expectations set by other applications
  - With some modifiers, scroll and zoom are combined in unexpected ways
  - Modifiers in the Graph Editor are not the same as typically used elsewhere

The current modifier + wheel behavior I am seeing is:
```
| Shift | Ctrl | Alt | Wheel | Scroll Behavior | Zoom Behavior |
|-------|------|-----|-------|-----------------|---------------|
|       |      |     | Up    | Right           | Vert-In       |
|       |      |     | Down  | Left            | Vert-Out      |
| X     |      |     | Up    | Up              | -             |
| X     |      |     | Down  | Down            | -             |
|       | X    |     | Up    | -               | Vert-In       |
|       | X    |     | Down  | -               | Vert-Out      |
| X     | X    |     | Up    | Up              | -             |
| X     | X    |     | Down  | Down            | -             |
|       |      | X   | Up    | -               | -             |
|       |      | X   | Down  | -               | -             |
| X     |      | X   | Up    | Right (slow)    | -             |
| X     |      | X   | Down  | Left (slow)     | -             |
|       | X    | X   | Up    | Right (fast)    | -             |
|       | X    | X   | Down  | Left (fast)     | -             |
| X     | X    | X   | Up    | Right (slow)    | -             |
| X     | X    | X   | Down  | Left (slow)     | -             |
```

I would recommend this behavior instead, to improve discoverability and be consistent with the majority of other software users are accustomed to:
```
| Shift | Ctrl | Alt | Wheel | Scroll Behavior | Zoom Behavior |
|-------|------|-----|-------|-----------------|---------------|
|       |      |     | Up    | Left            | -             |
|       |      |     | Down  | Right           | -             |
| X     |      |     | Up    | Up              | -             |
| X     |      |     | Down  | Down            | -             |
|       | X    |     | Up    | -               | Horiz-In      |
|       | X    |     | Down  | -               | Horiz-Out     |
| X     | X    |     | Up    | -               | Vert-In       |
| X     | X    |     | Down  | -               | Vert-Out      |
[...]

PR #564, opened 2019-02-27
Scrolling consistency improvements

mdmayfield:
### Issues addressed
- Fix left = right bug, and allow two-finger horizontal scrolling, as discussed in #505
- Implement #549
- Address #550
- Refactor code

### Specific Changes
- New option, "Horizontal Scroll Wheel (Shift toggles)"
  - True by default, for benefit of mouse users
  - Wording makes it explicit that Shift allows opposite-axis scroll
  - Turn this option off for a natural 2D-scrolling touchpad experience
- Reword option "Scroll Wheel Zooms" to add "(Ctrl toggles)"
  - Improve discoverability

### Timeline Changes
- Implement speed-sensitive scrolling and zooming for touchpads
  - This may make pixel scrolling unnecessary? Need further testing/research
- On a touchpad, 2D free scrolling now works as expected (when "Horizontal Scroll Wheel" is off)
- Make zooming in and out symmetrical (was 0.75 / 1.5, now 0.75 / 1.33)
- Either Shift or Alt will flip horizontal/vertical axis
- Zooming remains unaffected (vertical scroll for zoom always available)

### Graph View Changes
- Implement speed-sensitive scrolling and zooming for touchpads
  - This may make pixel scrolling unnecessary? Need further testing/research
- On a touchpad, 2D free scrolling now works as expected
- On a touchpad, multi-axis 2D free zooming now works as expected
- Honor "Scroll Wheel Zooms" and "Horizontal Scroll Wheel" settings
- Alt-scroll zooms both axes evenly (former scroll-wheel-alone behavior)
- Ctrl toggles zoom vs. scroll
- Shift toggles horizontal vs. vertical (for mouse wheels)
- Scroll / zoom behavior matches timeline

### DIFF
--- a/io/config.cpp
+++ b/io/config.cpp
@@ -70,7 +70,8 @@ Config::Config()
     center_timeline_timecodes(true),
     waveform_resolution(64),
     thumbnail_resolution(120),
-    add_default_effects_to_clips(true)
+    add_default_effects_to_clips(true),
+    horizontal_timeline_scroll(true)
 {}
 
 void Config::load(QString path) {
@@ -87,6 +88,9 @@ void Config::load(QString path) {
         } else if (stream.name() == "ScrollZooms") {
           stream.readNext();
           scroll_zooms = (stream.text() == "1");
+        } else if (stream.name() == "HorizontalTimelineScroll") {
+          stream.readNext();
+          horizontal_timeline_scroll = (stream.text() == "1");
         } else if (stream.name() == "EditToolSelectsLinks") {
           stream.readNext();
           edit_tool_selects_links = (stream.text() == "1");
@@ -233,6 +237,7 @@ void Config::save(QString path) {
   stream.writeTextElement("Version", QString::number(olive::kSaveVersion));
   stream.writeTextElement("ShowTrackLines", QString::number(show_track_lines));
   stream.writeTextElement("ScrollZooms", QString::number(scroll_zooms));
+  stream.writeTextElement("HorizontalTimelineScroll", QString::number(horizontal_timeline_scroll));
   stream.writeTextElement("EditToolSelectsLinks", QString::number(edit_tool_selects_links));
   stream.writeTextElement("EditToolAlsoSeeks", QString::number(edit_tool_also_seeks));
   stream.writeTextElement("SelectAlsoSeeks", QString::number(select_also_seeks));
--- a/io/config.h
+++ b/io/config.h
@@ -165,6 +165,7 @@ struct Config {
    * @brief The scroll wheel zooms rather than scrolls
    *
    * **TRUE** if the scroll wheel should zoom in and out rather than scroll up and down.
+   * The Control key temporarily toggles this setting.
    */
   bool scroll_zooms;
 
@@ -508,6 +509,14 @@ struct Config {
    * VolumeEffect, and PanEffect) added to them by default.
    */
   bool add_default_effects_to_clips;
+
+  /**
+   * @brief Horizontal timeline scroll
+   *
+   * **TRUE** Scrolling vertically with a mouse wheel or touchpad scrolls the Timeline horizontally.
+   * The Shift key temporarily toggles this setting.
+   */
+  bool horizontal_timeline_scroll;
 
   /**
    * @brief Load config from file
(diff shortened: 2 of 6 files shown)


elsandosgrande:
Seems like Macs work differently. Go figure.

mdmayfield:
@elsandosgrande What are you referring to exactly - the Travis build, or is there an issue with the scrolling/zooming on macOS? I have tested this patch on Ubuntu and Windows 10, and have a Mac but haven't compiled Olive on there yet.

alcomposer:
Can the `scroll wheel` used as right, left scroll not be a choice? I am used to pressing shift (even on a mouse) to scroll left and right. At least I think it should be a preference?  Other than that this looks like a very nice addition.

elsandosgrande:
@mdmayfield Travis-CI.

mdmayfield:
@alcomposer "Horizontal Scroll Wheel" is an option in this patch; or did you mean moving it to the Preferences dialog? Un-check Tools -> Horizontal Scroll Wheel and you should get exactly the behavior your describe on a laptop touchpad.

Honestly, my personal preference is the same as yours - if the choice were entirely mine, I wouldn't even bother including those options, and make the behavior strictly up/down for a mouse wheel, unless holding Shift; and always scrolling, unless holding Control.

elsandosgrande:
I've seen one other pull request before erroring out in macOS. Is this acceptable?

alcomposer:
@mdmayfield sorry, didn't notice the setting in tools.

itsmattkc:
Giving it some thought, I don't think any other panel should scroll horizontally apart from the timeline. The timeline's usage is primarily horizontal which isn't necessarily the case for the other panels. I think a better solution would be normal axes for everything, with the following option in Preferences > Behavior on by default:

```
[x] Invert Timeline scroll axes
```

I've been planning to move a number of those Tools menu behavior settings there eventually anyway.

Otherwise, everything looks good feels better than the existing scrolling in the master.

alcomposer:
As long as it's a preference I like it! Hey, this could be Olives version of left-click-select. ;-)

mdmayfield:
@itsmattkc Sounds great! I'll rebase on latest master and update the PR branch.

mdmayfield:
OK, I pushed a new version of this and the PR is updated! This ought to close #505, #549, and #550.

mdmayfield:
BTW these force-pushes are just to rebase on latest upstream and avoid a manual merge if/when this is pulled; there have been no changes to the actual patch since Friday.

itsmattkc:
Hi sorry it's taken so long for me to get around to this again. After some more testing, I have some more thoughts on the GraphView scroll.

I'm think Ctrl+Scroll should zoom uniformly on the GraphView and Ctrl+Alt+Scroll (or something) should zoom just one axis. For the benefit of mouse users, I'm thinking they'd rather zoom both axes by default.

Unfortunately as I've learned, and I'm sure you've noticed, Qt's scrolling behavior when holding Alt is actually not just switching axes, it's combining both deltas into the horizontal axis. I haven't seen any obvious way to override this behavior either so I'm not sure exactly how this would work, but I'm happy to do some research to see how this behavior can be achieved if this sounds like an okay idea.

mdmayfield:
@itsmattkc That's odd; I'm definitely seeing Alt only switch axes by default, and not combine them. The combination (Alt + any axis to zoom both axes) is something included in my patch to enable uniform zooming; maybe that's what you're seeing? Or I might be overlooking something.

On my laptop at least, with vanilla Qt scrolling and no axis swap shenanigans patched in, holding Alt makes touchpad left/right scroll up/down, and up/down scroll left/right; no combining, a straight-up swap.

Otherwise, it should be easy to make zoom both axes by default. I'm thinking:

Ctrl-Scroll up/down/left/right, zoom both axes
Ctrl-Shift-scroll up/down/left/right, zoom horizontally
Ctrl-Alt-scroll up/down/left/right, zoom vertically
Ctrl-Alt-Shift-scroll up/down/left/right, zoom both axes

That way all bases should be covered regardless of what Qt does. I will take a look at that tomorrow. Thanks!

itsmattkc:
That is really weird, it's definitely combining the QWheelEvent deltas specifically on my end, but maybe that behavior is exclusive to Qt on Windows (it also doesn't produce any pixelDelta even with a Windows Precision Touchpad so it wouldn't surprise me if it was a little underdeveloped). If the Alt works correctly on other platforms, we have no problems! Your plan sounds good.

mdmayfield:
@itsmattkc I bet you're right that it's a Windows Qt thing. My Windows box is a desktop with a mouse (so I would never have seen this behavior) and Linux is a laptop.

I think this patch is done; a quick test works as expected. Thanks!

PR #577, opened 2019-03-01
readme and issue template update

naj59:
Added a screenshot for new users, travis build status (with link) and in the issue template the amount of RAM the user has (if something like *no more memory*).


BrimsonBhin:
Perhaps add the Discord server too? :)

PR #582, opened 2019-03-02
Translations

elsandosgrande:
I think this is finally acceptable, right?


itsmattkc:
Yep looks good, thanks!

PR #591, opened 2019-03-03
Preliminary Haiku port

mmuman:
This makes it run on Haiku, although it crashes on exit and can't use OpenGL yet.
I think some ifdefs could use some cleanup, like qmake should pass a HAVE_DLOPEN or something instead.

### DIFF
--- a/effects/internal/vsthost.cpp
+++ b/effects/internal/vsthost.cpp
@@ -146,7 +146,7 @@ void VSTHost::loadPlugin() {
 #ifdef _WIN32
     DWORD dll_err = GetLastError();
     dll_error = QString::number(dll_err);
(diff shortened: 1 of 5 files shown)


mmuman:
[Looks good](https://twitter.com/mmu_man/status/1102293229381271552) at least.

elsandosgrande:
Well, good luck with adding support! I wonder if this makes Olive Haiku's first ever video editor.

mmuman:
No, we do have some native apps, but someone mentioned it on our forum, so I had a try.

itsmattkc:
That's a shame Haiku doesn't have QtOpenGL yet, but it's great seeing Olive running on a new OS!

PR #605, opened 2019-03-06
previously imported interlaced videos being shown/treated as progressive on subsequent imports

jonno85uk:
If a preview already exists a waveform was not generated and stream info (at least interlacing) was not found

(cherry picked from commit 70ef8cedcb7eba7a29b55563a03d10d5dd01a20a)

---

On the 1st import of an interlaced video, it's field order is correctly shown in the Media hover-popup and treated as an interlaced video. 
On the next import, the same interlaced video is shown as progressive, and is treated as such.

Moving PreviewGenerator::generate_waveform() out of the PreviewGenerator::retrieve_preview() condition allows the interlaced video to be treated correctly, but i'm not entirely sure why.  I saw Lines 106 to 108 for the reason.

On a slightly different note. Are interlaced videos actually working in Olive? It always shows either garbage (looks like empty frame data) or the last frame of the last shown video. At least in the preview viewer. I came across this issue whilst going through the caching thread in Chestnut.

### DIFF
--- a/io/previewgenerator.cpp
+++ b/io/previewgenerator.cpp
@@ -118,7 +118,7 @@ void PreviewGenerator::parse_media() {
 
       if (append) {
         QVector<FootageStream>& stream_list = (fmt_ctx_->streams[i]->codecpar->codec_type == AVMEDIA_TYPE_AUDIO) ?
-              footage_->audio_tracks : footage_->video_tracks;
+                                                footage_->audio_tracks : footage_->video_tracks;
 
         for (int j=0;j<stream_list.size();j++) {
           if (stream_list.at(j).file_index == i) {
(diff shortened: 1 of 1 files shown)


jonno85uk:
:headbash: whitespace changes hides the minor change

itsmattkc:
Sorry it's taken me this long to get to pull requests, it does appear interlaced video is broken at the moment. Probably something to do with the recent caching rewrite (or perhaps something even earlier, I haven't really tested it since it was first implemented).

I will firstly say, looking at it again, PreviewGenerator's functions are horribly named and very unintuitively written. I will definitely fix that soon. PreviewGenerator::retrieve_preview() inexplicably returns true if it **_failed_** to retrieve any previews and requires PreviewGenerator::generate_waveform() to run. Therefore placing PreviewGenerator::generate_waveform() outside PreviewGenerator::retrieve_preview() is probably not a good idea since one of its main functions is determining whether PreviewGenerator::generate_waveform() should be run in the first place.

The interlacing bug is quite logical however, yet something I hadn't picked up on before. FFmpeg's stream/codec data doesn't contain any information about whether the video is interlaced or not. The only way to determine that is to read a frame, which *does* have that information (AVFrame::interlaced_frame). So when PreviewGenerator skips generating the waveform, it also skips retrieving any information about interlacing. I have thought about caching the Footage/FootageStream metadata to file along with the thumbnails and waveforms (potentially negating having to run PreviewGenerator at all on re-imported media). This may be a good reason to do that.

jonno85uk:
Hi @itsmattkc ,

I can confirm earlier version of the cacher does work with interlaced files. 

I was afraid that moving generate_waveform() out of the retrieve_preview()  would cause unintended side-effects so I wasn't sure if it was the right thing to do. 

I have a junky email (no idea how to do PMs in github, which is kind of irritating) you can send a message to regomodo at gmail dot com

PR #609, opened 2019-03-06
Platform header, switching to Q_OS_

naj59:
This could probably help us in future. The general ``Q_xx_`` thing is very interesting.


eszlari:
I don't see the point in an extra level of defines.

As a Qt application, I think Olive should just use:
* `Q_OS_WIN` for Windows
* `Q_OS_MAC` for macOS
* `Q_OS_UNIX && !Q_OS_MAC` for Linux/BSD

itsmattkc:
I'm inclined to agree we may as well just use the Qt ones (be01bdd6f901bb88c37adcf5026f9c378456dc72). I know that's technically less "portable" if Olive were to ever switch away from Qt, but I think there would be far bigger concerns in that situation than these. I appreciate the suggestion though!

PR #630, opened 2019-03-14
add cmake buildsystem

eszlari:
advantages:
* easy to add new dependencies(+checking)
* fast with ninja back-end (`cmake -G Ninja`): https://ninja-build.org
* a lot of tools support it much better than qmake
* Qt 6 will be using cmake: https://lists.qt-project.org/pipermail/development/2018-October/034023.html
* clang-tidy integration: https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_CLANG_TIDY.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,479 @@
+cmake_minimum_required(VERSION 3.8 FATAL_ERROR)
+
+project(olive-editor LANGUAGES CXX)
+
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+set(CMAKE_AUTOMOC ON)
+set(CMAKE_AUTOUIC ON)
+set(CMAKE_AUTORCC ON)
+
+set(OLIVE_DEFINITIONS -DQT_DEPRECATED_WARNINGS)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+
+if(UNIX AND NOT APPLE AND NOT DEFINED OpenGL_GL_PREFERENCE)
+  set(OpenGL_GL_PREFERENCE GLVND)
+endif()
+find_package(OpenGL REQUIRED)
+
+find_package(Qt5 5.7 REQUIRED
+  COMPONENTS
+  Core
+  Gui
+  Widgets
+  Multimedia
+  OpenGL
+  Svg
+  LinguistTools
+)
+
+find_package(FFMPEG 3.4 REQUIRED
+  COMPONENTS
+  avutil
+  avcodec
+  avformat
+  avfilter
+  swscale
+  swresample
+)
+
+find_package(frei0r)
+if(NOT FREI0R_FOUND)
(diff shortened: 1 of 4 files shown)


eszlari:
I have only tested this on Linux so far.

DeveloperPaul123:
I can test this out on windows, but just looking at the CMakeLists I would suggest not using `PkgConfig` for windows. Something like `vcpkg` would play nicer on windows with cmake. Especially with visual studio.

@eszlari Just curious but what distribution of Linux did you test this on?

itsmattkc:
`pkgconfig` would work if the user is compiling with MinGW. Perhaps have an option for that or `vcpkg` (and/or automatically check whether the build environment is MinGW or MSVC).

eszlari:
> but just looking at the CMakeLists I would suggest not using PkgConfig for windows.

I removed pkgconfig usage.

> Just curious but what distribution of Linux did you test this on?

Kubuntu 18.10

DeveloperPaul123:
@eszlari I had to make some minor changes to allow for compilation to work. I also had to make a minor code change to fix a compilation error regarding the use of `LoadLibrary()` on windows. Should I create a PR to your `cmake` branch or just commit them to the existing branch you have?

eszlari:
I would open a PR against upstream for the compile fix  and a PR against my cmake branch for your MSVC support.

DeveloperPaul123:
@eszlari Ok will do

DeveloperPaul123:
@eszlari I filed a PR against your `cmake` branch.

DeveloperPaul123:
Is there any desire to add installation support and possible installer generation support via CPack?

eszlari:
@itsmattkc This is more or less ready from my side, any comments?

eszlari:
Small speed comparison:

qmake+make:
```
real    1m0,535s
user    7m21,124s
sys     0m27,312s
```
cmake+ninja:
```
real    0m35,959s
user    4m6,466s
sys     0m15,528s
```

itsmattkc:
Very interesting, give me some time to test it properly and I'll get back to you within a few days

DeveloperPaul123:
Since this is taking some time to merge I'd like to propose adding in auto building (if needed) of some of the optional dependencies of olive to take full advantage of CMake.

Specifically I had `frei0r` and `OpenColorIO` in mind. Thoughts?

itsmattkc:
Apologies for the late merge, everything seems good.

Auto-building may be good one day for OpenColorIO, availability for it seems to vary a lot per platform/distribution. Don't bother with Frei0r though, [it's getting phased out](https://github.com/olive-editor/olive/wiki/Frei0r-Support-EOL) and either way doesn't really need to be built for Olive, it only requires including `frei0r.h` to compile support in.

DeveloperPaul123:
Sounds good, a PR for OpenColorIO would be relatively easy to do. Another option is to build OpenColorIO as part of `olive` using CMake's [`FetchContent`](https://cmake.org/cmake/help/v3.11/module/FetchContent.html) as opposed to [`ExternalProject`](https://cmake.org/cmake/help/v3.11/module/ExternalProject.html?highlight=externalproject). Is there a preference?

sobotka:
Chaining #826 into this.

PR #639, opened 2019-03-17
Automated cut based on sound

morrolinux:
This feature enables you to automatically cut silences in videos containing speeches such as screencasts etc. One can adjust attack and release threshold and time for best results. However, default settings should be fine for a standard speaker (non whispering) in a relatively quiet room.

TODO: fix the ComboAction pushing such that one CTRL+Z stroke can revert the entire thing rather than just one cut at a time

olive audio cut with dialog

### DIFF
--- a/dialogs/silencedialog.cpp
+++ b/dialogs/silencedialog.cpp
@@ -1,0 +1,192 @@
+/***
+
+    Olive - Non-Linear Video Editor
+    Copyright (C) 2019  Olive Team
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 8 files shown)


alcomposer:
Putting this into a *Tools* menu in the sequence contextual menu would be very cool. I can see many more tools like this being added.

albacoretuna:
Is this feature removed from the latest nightlies?

Simran-B:
v0.2 is a rewrite and it looks like this feature wasn't ported from v0.1. It would have to be implemented differently because there is a playback cache now.

itsmattkc:
Yeah you may see it reimplemented in 0.2, but it'd have to be implemented very differently and it'd probably be better as a script of some kind if we implement scripting functionality.

PR #678, opened 2019-03-23
Add Indonesian translation

ZoomTen:
Adds support for `Bahasa Indonesia` in Olive.
You could probably ignore the commit that modifies `olive.pro`, I used it to test my translation.

Is the autosave dialog, etc. translatable now? Can't seem to find it when I updated my `ts` file.

### DIFF
--- a/olive.pro
+++ b/olive.pro
@@ -316,7 +316,8 @@ TRANSLATIONS += \
     ts/olive_ru.ts \
     ts/olive_uk.ts \
     ts/olive_bs.ts \
-    ts/olive_sr.ts
+    ts/olive_sr.ts \
+    ts/olive_id.ts
 
 win32 {
     RC_FILE = packaging/windows/resources.rc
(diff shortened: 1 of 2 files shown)


itsmattkc:
Thanks for the translation! Sorry, which auto-save dialog?

ZoomTen:
@itsmattkc The one that shows up upon restarting after a crash, that says "Olive has found a recovery file, would you like to open it?"

There might be a couple of more things that have missing translatable strings

itsmattkc:
@ZoomTen Strange, the string is wrapped in `tr()` here: https://github.com/olive-editor/olive/blob/master/global/global.cpp#L97

Maybe it doesn't like the fact that the string is split across two lines.

ZoomTen:
@itsmattkc Ah, yeah. I checked my `ts` file again and I *did* translate it, my mistake.

It might also be that the locale hasn't been loaded by the time the message showed up, I use a custom stylesheet and the message showed up in the usual "Fusion" style.

Issue #698, opened 2019-03-26
Appimage segfault in Debain 10 Buster SID


cgvirus:
### System Information

**Olive version:**  Olive-675ea94-x86_64.appimage
**Operating system:** Debain 10 Buster (SID)

### Detailed Description

Doesn't open.
Segmentation fault
Although building from source works. 

### Steps to reproduce

Appimage opening with terminal

### Output Log

2019-03-27T00:47:36 [INFO] Initializing custom cursors
2019-03-27T00:47:36 [INFO] Finished initializing custom cursors
2019-03-27T00:47:36 [INFO] Initializing icons
2019-03-27T00:47:36 [INFO] Finished initializing icons
2019-03-27T00:47:37 [INFO] Initializing effects...
Segmentation fault

AGI-THB:
can you post a stacktrace

cgvirus:
Hi building from source is working. 
The appimage in Github is giving this error:

2019-03-27T00:47:36 [INFO] Initializing custom cursors
2019-03-27T00:47:36 [INFO] Finished initializing custom cursors
2019-03-27T00:47:36 [INFO] Initializing icons
2019-03-27T00:47:36 [INFO] Finished initializing icons
2019-03-27T00:47:37 [INFO] Initializing effects...
Segmentation fault

oc1024:
https://github.com/olive-editor/olive/wiki/How-to-get-output-logs-if-the-program-crashes-and-use-GDB-LLDB-for-more-detailed-logshttps://github.com/olive-editor/olive/wiki/How-to-get-output-logs-if-the-program-crashes-and-use-GDB-LLDB-for-more-detailed-logs

Wiki page on how to get a backtrace.

cgvirus:
Hi with Build I get 
`Reading symbols from ./olive-editor...(no debugging symbols found)...done.`

But Building is working properly. No problem. Already done a small project as well.

Appimage is not working though. I am not sure how can I give you the stack for appimage as It doesn't start.

itsmattkc:
@cgvirus Running GDB on the AppImage should still provide a trace up until the crash, however compiling yourself probably provides a better experience with Olive anyway.

cgvirus:
Hi this what it says:
```
Reading symbols from ./Olive-675ea94-x86_64.AppImage...(no debugging symbols found)...done.
(gdb) 
```

itsmattkc:
[...]

PR #709, opened 2019-03-28
Three types of Glitch effect Added

cgvirus:
Hello my first contrib for Effects in Olive. There are 3 types of Glitch effects in distort section.

### DIFF
--- a/effects/shaders/glitch.frag
+++ b/effects/shaders/glitch.frag
@@ -1,0 +1,269 @@
+uniform sampler2D tex;
+varying vec2 vTexCoord;
+uniform float time;
+uniform vec2 resolution;
+
+uniform float THRESHOLD;
+
+uniform int mode;
+
+#define PI 3.14159265
+#define TILE_SIZE 16.0
+
+
+float sat( float t ) {
+	return clamp( t, 0.0, 1.0 );
+}
+
+vec2 sat( vec2 t ) {
+	return clamp( t, 0.0, 1.0 );
+}
+
+//remaps inteval [a;b] to [0;1]
+float remap  ( float t, float a, float b ) {
+	return sat( (t - a) / (b - a) );
+}
+
+//note: /\ t=[0;0.5;1], y=[0;1;0]
+float linterp( float t ) {
+	return sat( 1.0 - abs( 2.0*t - 1.0 ) );
+}
+
+
+//note: [0;1]
+float rand( vec2 n ) {
+  return fract(sin(dot(n.xy, vec2(12.9898, 78.233)))* 43758.5453);
+}
+
+float rand(float n)
+{
+    return fract(sin(n) * 43758.5453123);
+}
+
+float noise(float p)
+{
+    float fl = floor(p);
+  	float fc = fract(p);
+    return mix(rand(fl), rand(fl + 1.0), fc);
+}
+
+float noise(vec2 p)
+{
+    vec2 ip = floor(p);
+    vec2 u = fract(p);
+    u = u * u * (3.0 - 2.0 * u);
+
+    float res = mix(
+        mix(rand(ip), rand(ip + vec2(1.0, 0.0)), u.x),
+        mix(rand(ip + vec2(0.0,1.0)), rand(ip + vec2(1.0,1.0)), u.x), u.y);
+    return res * res;
+}
+
+
+//note: [-1;1]
+float srand( vec2 n ) {
+	return rand(n) * 2.0 - 1.0;
+}
+
+float trunc( float x, float num_levels )
+{
+	return floor(x*num_levels) / num_levels;
+}
+vec2 trunc( vec2 x, float num_levels )
+{
+	return floor(x*num_levels) / num_levels;
+}
+vec2 trunc( vec2 x, vec2 num_levels )
+{
+	return floor(x*num_levels) / num_levels;
+}
+
+
+
+vec3 spectrum_offset( float t ) {
+	vec3 ret;
+	float lo = step(t,0.5);
+	float hi = 1.0-lo;
+	float w = linterp( remap( t, 1.0/6.0, 5.0/6.0 ) );
+	float neg_w = 1.0-w;
+	ret = vec3(lo,1.0,hi) * vec3(neg_w, w, neg_w);
+	return ret;
+}
+
+vec3 rgb2yuv( vec3 rgb )
+{
+	vec3 yuv;
+	yuv.x = dot( rgb, vec3(0.299,0.587,0.114) );
+	yuv.y = dot( rgb, vec3(-0.14713, -0.28886, 0.436) );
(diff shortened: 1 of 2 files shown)


NathanSalapat:
I'm a nobody in this project, but you might get better luck with some pictures or explanations of what the glitch effects are. :)

BrimsonBhin:
I do not know if this effect should be in the `distort` category. Distort effects in olive are mathematics that do not tamper with the color of the video. This one enforces some color changes for the aesthetic. Here are the 3 effects

**Signal Loss**
signal loss

**Bad TV**
bad tv

**Worse TV**
worse tv

Is there a way to configure the blockiness and direction of the effect? That would be perfect. If you have way more, I think you can create your own repo for you own custom effects so Olive does not become messy. I would gladly star that repo and link it in the Wiki :).

BrimsonBhin:
This seems to be a port of [this shader form shadertoy](https://www.shadertoy.com/view/lsfGD2).

Unless that Shadertoy account is your's, I would also advise giving credit

alcomposer:
Shader toy may / may not be GPL depended on the license that is associated with the effect. And if so yes the license needs to remain in the shader source.

Either way, effects like this don't need to be added to Olive itself, they can be imported by the user very easily.

BrimsonBhin:
Yup, a separate repository with a README on how to add them would suffice.

alcomposer:
My motivation for pointing this out is that @itsmattkc has already implemented a very advanced GLSL plugin system, and that there could easily be 1,000,000 effects users could want to use. Other software has effects plugins that are not part of the software. (Even NUKE has external OpenFX effects).

cgvirus:
Hi folks Thanks for starting the conversation.

- Licence: I almost forgot about the licence block. When I was learning GLSL these three effects were my part of research and I always start with these to make sure all uniforms are under API compliance. It seems Olive has all Uniform types under the hood and now my template is ready for future.
I will attach the licence block here which are Unlicensed and GPL. Will link them in the block.

- Category: I agree with the point of Distort. I think a dedicated GPU Effect section would be better.

- **Repository:** PLEASE Let me know if this can be done! I am searching for it! If we can only add this effects by Olive preference path or somehow then I don't have to push/pull and wait for Olive to be build (Which can become pretty huge BTW). I can control my decisions and corrections too. And Worldwide contribs can merge there effects easily as well. 
How can we do it specially with Appimage? I have some of my shaders and vert scripts and the good thing is Olive already supports a pretty huge GLSL arena. So there is a huge area Olive can cover itself.

I am glad the conversation has began. The earlier is the better.
Let me know folks!

@alcomposer @BrimsonBhin @itsmattkc

alcomposer:
At this stage it could be as simple as hosting your effects in your own repository on GitHub. N need to have the rest of Olive there, but have a dedicated repo for your effects.

Currently there are many huge changes happening to Olive, least of which is transitioning to OCIO color management, and using Scene Linear working space. This will effect GLSL effects that deal with color, as they will need to be able to accept, and output in scene linear. Any sort of effect that needs to invert the color values for example (1-c) will fail, as scene linear is not clamped from a 0<1 range.

As @BrimsonBhin has pointed out, having a folder, with a readme at this point would enable users to work out how to use the GLSL filters.

Remember that as Olive is still very young, and currently going though huge changes, there may be further changes to come in the effects infrastructure. IIRC @itsmattkc has needed to move to a new OpenGL version to allow OCIO integration. He probably would be the best to comment on that.

Other than that, that you for making these GLSL effects work with Olive, and I hope to see more in the future!  :+1:

cgvirus:
@alcomposer Alright, So if I create a repo with effects/shader folder then how can users (I mean average users) can install them?

Currently I am doing it by building Olive in my Debian box. As appimage has some issue #698

Here I have found a folder effects/shader. And I am following the xml to load it.

Please let me know. Thanks. 
I would like to promote Olive within this power of GLSL and my audience most of them can't understand complication. I will make a README with this instruction as well. Thanks for your time.

alcomposer:
> Alright, So if I create a repo with effects/shader folder then how can users (I mean average users) can install them?

Depending on your target audience, you could always include a simple installer with your shaders? All it would need to do is know the installed directory of Olive, and then place all the files into the shaders folder? That may work?

cgvirus:
Yes, that may work but Appimage comes only as a single pack. I think if we can have a path option in preference to link the shaders it may work. @alcomposer

itsmattkc:
>Category: I agree with the point of Distort. I think a dedicated GPU Effect section would be better.

Keep in mind that *all* of Olive's effects are GPU effects (with the exception of Frei0r which is currently being phased out) so a "GPU Effect" section is unnecessary. I agree with @BrimsonBhin that "Stylize" is the correct place to put this.

Anyway when it comes to merging these, they look good and I am interested in merging them, but there are certainly concerns regarding licensing and scene linear. I'm not sure how Shadertoy works (i.e. if you need to contact the authors or there's a stock license, etc.) and I see some clamps between 0.0 and 1.0 which may be problematic. They'll have to be tested with the `furtherocio` branch.

cgvirus:
@itsmattkc Thanks. I will update the licence and let you know. But rather than merging can we make a search path system in Olive preference? So that, users can load or write their own shaders and test it. It would also help the devs like us to share at maximum speed with users and also with Olive Team.

itsmattkc:
I'm not really convinced having an extra search path that users have to manage is any easier than just copying files to `/usr/local/share/olive-editor/effects`. A local user path like `~/.local/share/olive-editor/effects` may be a good idea for those without root access though. For developers there is the `OLIVE_EFFECTS_PATH` environment variable which will add an extra search path.

cgvirus:
So what should I say to the users using appimage? I have tried this link copying effects.
`/usr/local/share/olive-editor/effects`
but appimage can't link it. 
Which package user should use in this case as Debian package is too old and buggy.

Let me know please all the links user should copy the effects into (Win, Mac, OSX). I will create a repo only with shaders so that Olive Team can focus on the main app. I will make README with your instructions.

Thanks. 
@itsmattkc

cgvirus:
Hi regarding the topic of path:
issue ##721

cgvirus:
Hi I have created a repo here:
https://github.com/cgvirus/Olive-Editor-Community-Effects

I am closing this merge as the recent changes https://github.com/olive-editor/olive/commit/a581d36bcc830bf81bcb2526596b0d0ff44e7768 is ready for personal contribution.

Please feel free to contribute in this repo and may be this can become a vault for Olive-Editor community. Let me know if you want to join in this repo for some admin privilege as well. I will add more effects soon.

Thanks @BrimsonBhin @alcomposer @itsmattkc

cgvirus:
Also @itsmattkc can we make subcategories through xml? like community>stylized>Glitch? It may become over-flooded soon so a subcategory would be nice I think. Thanks.

PR #731, opened 2019-03-30
Italian localization

MassiminoilTrace:
- Added italian localization
- I ran `lupdate` to update and generate new lines to be translated. Seems that the previous `lupdate` was made some time ago.
Should I also run `lrelease` or is it automatic?


itsmattkc:
Yes the files in the branch are not `lupdate`d automatically. `lrelease` is unnecessary as that's performed at build time.

MassiminoilTrace:
Ok, I have already run lupdate for other languages also and included these changes in the commit :)
Anyway, I've just noticed that I need to refine some translations, please do not merge this pull request now. I write here when I finish. I should finish it in few hours.
Thank you for your previous answer!

MassiminoilTrace:
Ok, now it's correct! My changes can be merged.

itsmattkc:
It's actually preferable to only update/commit the language file you're working on since committing all of them will cause merge issues if anyone else is working on the languages at that time. Anyway thank you very much for the translation! I've merged it in in 1a12475e00d81b3aaa76c564e6b6cc36f652704f

PR #759, opened 2019-04-05
cmake: add doxygen

eszlari:


itsmattkc:
Thanks for the PR, though I would prefer a solution that didn't remove the Doxyfile as I may want to generate the docs without re-building (unless there's a way to cmake just the docs?)

eszlari:
You would just run:
```
$ cmake -GNinja -DBUILD_DOXYGEN=ON ..
$ ninja docs
```

eszlari:
The line 
```
doxygen_add_docs(docs ALL ${OLIVE_SOURCES})
```
adds the custom build target `docs`.

itsmattkc:
I have a script on the webserver that uses doxygen, and I was only barely able to install doxygen on it let alone cmake and ninja. Is there no way this can just offload to the doxyfile?

eszlari:
>  I was only barely able to install doxygen on it let alone cmake and ninja.

I don't understand. What is hard about installing one of the most common build systems around?

itsmattkc:
I have an auto-build script for the documentation on a shared webserver that I don't administer - I don't have readily available access for installing software. Doxygen could be uploaded a single static binary, but I have doubts cmake/ninja would be as easy (if possible).

eszlari:
> I have an auto-build script for the documentation on a shared webserver that I don't administer - I don't have readily available access for installing software.

Can't you ask the admin to install cmake/ninja?

> Doxygen could be uploaded a single static binary, but I have doubts cmake/ninja would be as easy (if possible).

https://github.com/Kitware/CMake/releases/download/v3.14.3/cmake-3.14.3-Linux-x86_64.tar.gz
https://github.com/ninja-build/ninja/releases/download/v1.9.0/ninja-linux.zip

itsmattkc:
What does removing Doxyfile improve that necessitates doing so/changing the existing workflow? How is deferring to an external Doxyfile in the CMakeLists.txt objectively worse?

eszlari:
> What does removing Doxyfile improve that necessitates doing so/changing the existing workflow?

That's just how the doxygen macros in cmake work:
https://cmake.org/cmake/help/latest/module/FindDoxygen.html

>  How is deferring to an external Doxyfile in the CMakeLists.txt objectively worse?

As far as I can tell, you used a template that barely changes any default doxygen values, so all these lines are redundant. In the cmake file, you only need to add the values that you actually want to change.

itsmattkc:
I'll merge this, but keep a Doxyfile on the server for the auto-generation.

PR #803, opened 2019-04-14
lupdate and german translation update

naj59:


itsmattkc:
Hi thanks for the update, but could you restrict your PR to only the updated German translation? Updating the other TS files in the repository could cause merge issues if anyone else is working on one of them locally.

naj59:
Problem at this point is that without ``lupdate`` it refers to the old source points where no translation is going to be found.

itsmattkc:
Is that the case with the resulting `.qm` files?

naj59:
I don't understand this concretly. Without updating the ``.ts`` files old strings are "not used anymore". This would probably resist in a non-translated ui.

itsmattkc:
I suppose if anyone's having merge issues, they can just override the repo's version with theirs.

PR #800, opened 2019-04-14
Filetype icon update? (dark palette)

ZoomTen:
I proposed earlier on the Discord that Olive could use its' own file type icon. Here's my take on an icon design. Might be related to #429.

Was considering making a pull request for the [light icons](https://github.com/ZoomTen/olive/commit/b24f3e580fa5c2c549afa5b3d831bf8f179c3d27), but I liked the dark ones more. You might want to [cherrypick](https://gist.github.com/ozh/cbce675ba35e0d336cff) this PR.

[Source svg](https://github.com/olive-editor/olive/files/3077688/mimetype_icons.zip), in case anyone wants to improve on it.

screenshot443
Large icon

screenshot442
In Olive's file selection box

(This PR only applies to Linux packaging, I haven't figured out how filetype icons work on Windows yet - perhaps they are embedded into the executable?)

PR #862, opened 2019-04-21
Filetype icons (light palette, Windows & Linux)

ZoomTen:
Continuing from #800, here's my proposal for a file type icon. This time, the icons use a light palette.

I included Windows support by adding the icon into the resource file, as well as update the NSIS script to add the file association. Let me know if there's something wrong with them!

[Source SVG](https://github.com/olive-editor/olive/files/3100950/application-vnd.olive-project_2.zip)

screenshot466
screenshot473


eszlari:
Note: Linux also support SVG directly (installed under `share/icons/hicolor/scalable`).

ZoomTen:
@eszlari I'm aware; I'm following conventions since most applications seem store their hicolor icons as PNGs, including Olive. Plus, it might help to avoid problems with the blur effect being used in the SVG.

PR #872, opened 2019-04-23
Make the user able to change the default image clip duration

v0idzz:
This adds a setting in the preferences window for setting the default duration of an image, instead of using  the hardcoded '100' value.

### DIFF
--- a/dialogs/preferencesdialog.cpp
+++ b/dialogs/preferencesdialog.cpp
@@ -318,6 +318,9 @@ void PreferencesDialog::accept() {
     delete_previews(delete_match);
   }
 
+  // Save default image duration
+  olive::CurrentConfig.default_image_duration = default_image_duration_timeedit->time();
+
   // Save keyboard shortcuts
   for (int i=0;i<key_shortcut_fields.size();i++) {
     key_shortcut_fields.at(i)->set_action_shortcut();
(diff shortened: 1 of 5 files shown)


dialogs/preferencesdialog.cpp
```diff
@@ -572,6 +575,17 @@ void PreferencesDialog::setup_ui() {
 
   row++;
 
+  // General -> Default Image Duration
+  general_layout->addWidget(new QLabel(tr("Default Image Duration:"), this), row, 0);
+
+  default_image_duration_timeedit = new QTimeEdit(this);
+  default_image_duration_timeedit->setDisplayFormat("HH:mm:ss.zzz");
+  default_image_duration_timeedit->setMinimumTime(QTime(0, 0, 0, 1));
```
Simran-B: Are clips allowed to be only 1ms long?

dialogs/preferencesdialog.h
```diff
@@ -243,6 +244,11 @@ private slots:
    */
   QSpinBox* waveform_res_spinbox;
 
+  /**
+   * @brief UI widget for selecting the default length of an image footage
```
Simran-B: image footage -> image clip?

global/config.cpp
```diff
@@ -216,6 +217,9 @@ void Config::load(QString path) {
         } else if (stream.name() == "WaveformResolution") {
           stream.readNext();
           waveform_resolution = stream.text().toInt();
+        } else if (stream.name() == "DefaultImageDuration") {
+          stream.readNext();
+          default_image_duration = QTime::fromString(stream.text().toString(), olive::kDefaultTimeFormat);
```
Simran-B: Look like there are no guards against bad config values (not just here, but in general), e.g. a negative image default duration

dialogs/preferencesdialog.cpp
freshstudio: Seems so.

dialogs/preferencesdialog.h
freshstudio: Yeah, you're right.

global/config.cpp
freshstudio: Well, that's what I noticed as well. I think we need a better config handling in general.

freshstudio:
> There are conflicts, probably because of the pipeline merge.

Yes. I was told to wait a little before doing any changes. @itsmattkc is it OK if I start working on it?

Issue #789, opened 2019-04-12
Add an "Invert" option to the Crop Effect


unfa:
It'd be great if the Crop effect could work in reverse.

Instead of cutting a rectangle from the middle and rejecting the outer parts - I'd sometimes like to reject the inner parts and just keep the borders.

DaniSeeh:
This might ultimately be treading on Masking territory and already be planned to be handled via that avenue. Up to @itsmattkc on how easy this would be to implement and whether it is worth doing or just waiting to Masking naturally fills that demand.

frink:
If we're talking masks there should be good feathering to go with that...

unfa:
Looking at Nodes - possibly an Invert node (effect) could be added with option to select which channels should be inverted (R, G, B, A).

That'd make it much more useful.

cgvirus:
@unfa 
Have created a pull request. For now you can check it here:
https://github.com/cgvirus/olive/blob/master/effects/shaders/crop.frag
https://github.com/cgvirus/olive/blob/master/effects/shaders/crop.xml

itsmattkc:
I think a general "alpha invert" node in the node restructure would be the best long-term solution here

cgvirus:
Or better a node named invert where R G B or A can be inverted with bool.

itsmattkc:
Hmm I'm not sure there's such thing as inverted R/G/B in scene linear. @sobotka ?

sobotka:
@itsmattkc Ha! I had a long post to this very issue, then deleted it because I didn’t familiarize myself with the code.

You are indeed absolutely correct, and it is quite a wise observation. Even `max(RGB)` will have a shortcoming unless calculated every time. “Negative” is literally an output referred distillation.

Most of these sorts of slipper ideas can be more accurately solved if they are broken down to what they are trying to do. For vignette, thankfully, it’s quite easy: It constricts light near the periphery of the lens due to optics and construction.

So what does “invert” do? Is it inverting the light constricting in the centre region? Is it increasing the exposure in the vignette region?

[...]

PR #880, opened 2019-04-25
Crop and vignette Inverse Feature added

cgvirus:
response to:
https://github.com/olive-editor/olive/issues/789 &
https://github.com/olive-editor/olive/issues/881


unfa:
The Crop effect inverts Alpha channel of the entire thing, not just the Crop result itself. I think the Crop's alpha should be calculated seaprately and multiplied with the incoming Alpha for a correct result.

Screenshot_20190503_173208
Screenshot_20190503_173226

In the above example Crop fights wiht the Chroma Key. I can work this around and get the expected result by applygin Crop first before Chroma Key:

Screenshot_20190503_173354

I'd be for fixing the Crop and adding a "Alpha Invert" effect (I guess alpha will always be between 0 and 1 even in scene-reffered model, right, @sobotka ?) so that should make sense fot the Color-Managed Pipieline too I guess?

What do you think?

sobotka:
In an associated alpha encode, the RGBA should be convolved as one unit.

Regarding how alpha is combined / constructed, it would depend. There are technically more than a single case to deal with, depending on the underlying geometry assumptions. For example:
1. Geometry is correlated, or perfectly aligned. Imagine a triangle dividing the pixel in half, and another triangle perfectly snapping up against it.
2. Geometry is completely non-overlapping. Imagine the two triangles are present, but not overlapping.
3. Geometry is uncorrelated, arbitrary overlapping. The two triangles are overlapping.
4. Geometry is completely overlapping, and foreground is fully occluding.

This breaks down into variants of the generic over function `A + (1 - A) * B` or `A + B - A * B`. Nuke has several merge node math functions that describe the above situations via over, disjoint over, and conjoint over. In order, the alpha math works out to roughly, as best as I understand things:
1. `A + B`
2. `A + B`
3. `A + B - A * B`
4. `A`

Nuke’s disjoint:`if(a1+a2<=1, c2+c1, c2+c1*(1-a2)/a1)`
Nuke’s conjoint: `if(a2>a1, c2, c2+c1*(1 - a2/a1))`

Examples of correlated geometry would be a holdout matte, where you subtract a foreground object from a background element.

The generic over function is also the basic probability function.

00CF49CA-956C-434A-A5CA-DC27DE1073B1
F7CB23CA-3598-4479-8BD9-13F3C0FE4B67

cgvirus:
Let me check it should not be too much hard to do. Will PR soon.

cgvirus:
@unfa Let me know if it's fixed:
https://github.com/cgvirus/olive/blob/master/effects/shaders/crop.frag

PR: https://github.com/olive-editor/olive/pull/904

PR #886, opened 2019-04-28
Keying/Despill effect

unfa:
Useful for improving chroma-keyed footage. Can work for R, G or B channels.
Defaults to green screen.

I made this, becasue I needed it, but it'd be great if others could use that too.

I've tested this and it works - i'ts so simple it's hard to break anything too badly...

### DIFF
--- a/effects/internal/despill.frag
+++ b/effects/internal/despill.frag
@@ -1,0 +1,59 @@
+/***
+    Despill - for cleaning up keying residue.
+    Implementation by unfa for Olive-Editor Community
+***/
+
+uniform sampler2D tex;
+varying vec2 uTexCoord;
+uniform vec2 resolution;
+uniform int channel;
+
+const vec2 renderScale = vec2(1,1);
+
+void main(void)
+{
+    vec2 uv = gl_FragCoord.xy/resolution.xy;
+
+    vec4 col = texture2D(tex, uv);
+    
+    // separate channels
+    
+    float r = col.r;
+    float g = col.g;
+    float b = col.b;
+    float a = col.a;
+    
+    // Select despilled channel: 0 = RED; 1 = GREEN; 2 = BLUE;
+    
+    if (channel == 0) // RED
+    { 
+        // replace new green channel by avaraging between existing red and blue channels
+        float r2 = (g + b) * 0.5;
+        
+        // now composite the original and new gree channel using the "darken" mode
+        r = min(r, r2);
+    }
+    
+    if (channel == 1) // GREEN
+    { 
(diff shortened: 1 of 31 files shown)


unfa:
@sobotka  Alright, I've done everything I could, but for some reason the shader doesn't work now. Is there a way to compile this and get some errors? Currently I just save the files and restart olive to see if it works - if it doesn't I try to find mistakes blindly. I've found and fixed many already, but still it fails.

cgvirus:
@unfa You should use terminal to open Olive Editor to see the error trace. Not only for Olive but any standard program, whenever you need to code you should use terminal to log your errors. For Linux, open or `cd` your terminal in your installation directory, then:
`./olive-editor`
If you have already loaded the frags, after editing the frag just save it from text editor and reassign the effect deleting the old one. The frag will be updated. No need to restart Olive.

effects/internal/despill.frag
```diff
@@ -7,11 +7,15 @@ uniform sampler2D tex;
...
 uniform vec2 resolution;
 uniform int channel;
+uniform float factor;
 
 const vec2 renderScale = vec2(1,1);
 
 void main(void)
 {
+    
+    factor = factor * 0.01; // convert factor from 0-100 scale to 0-1 scale
```
alcomposer: Factor is a read only variable, you need to make a new `float factorOut` and use that from then on:

```
float factorOut;
factorOut = factor * 0.01;
```
Then replace all the other factors with factorOut

alcomposer:
I left a comment in your PR how to fix. Also why is this in `effects/internal`? Should it not be in `shaders`?
If you decided to chance the directory- also make sure to change the location of the COMMON.vert in the XML.

effects/internal/despill.frag
unfa: I see. Thanks! Fixed!

unfa:
> Why is this effect in `effects/internal`? Should it not be in `shaders`?
> If you decide to chance the directory- also make sure to change the location of the COMMON.vert in the XML.

You're right. I first used the Community-Effects asa template hence my mistake. Fixed.

It works now!

I also want to add a "Balance" factor to allow shifting the weighting of channels used to reproduce the missing one.

unfa:
I've pushed another version whht added Balance parameter and refactored code to use functions instead of repeating the same stuff 3 times.
Easier to debug, and more efficient.

I tested it - works as expected.

itsmattkc:
Merged in dff8db1af25a4ab3f0d580cf1c59fb2ffc3666e3

PR #887, opened 2019-04-29
Update coefficients from busted up REC.601

sobotka:
The Chromakey fragment shader is a complete mess, isn't colour managed, and
likely a terrible choice of colour encoding model for keying work, but I've updated
it anyways.

The existing coefficients are assuming the RGB colourspace as being the rather
batshit loopy REC.601 definition. This patch should update it to use the assumed
input of REC.709, as well as change the YCbCr values for REC.709 input.

No promises...

### DIFF
--- a/effects/shaders/chromakey.frag
+++ b/effects/shaders/chromakey.frag
@@ -22,7 +22,7 @@ float rgb2y (vec3 c) {
 // 0.2126 R, 0.7152 G, and 0.722 B. Easy change.
 // The coefficients for YCbCr are calculated off of the
 // REC.709 values.
-	return (0.2126*c.r + 0.7152*c.g + 0.0.722*c.b);
+	return (0.2126*c.r + 0.7152*c.g + 0.0722*c.b);
 } 
  
 float rgb2cb (vec3 c) { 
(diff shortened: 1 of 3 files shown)


effects/shaders/chromakey.frag
```diff
@@ -15,17 +15,24 @@ uniform int mode;
...
 	/*a utility function to convert colors in RGB into YCbCr*/ 
-	return (0.299*c.r + 0.587*c.g + 0.114*c.b);
+
+// This isn’t colour managed and is a huge mess, but in the
+// short term, using correct weights will give significantly
+// more ideal results. The correct weights for REC.709 are
+// 0.2126 R, 0.7152 G, and 0.722 B. Easy change.
+// The coefficients for YCbCr are calculated off of the
+// REC.709 values.
+	return (0.2126*c.r + 0.7152*c.g + 0.0.722*c.b);
```
BrimsonBhin: Simple typo. `0.0.722*c.b` should be `0.722*c.b` (otherwise the effect would not work)

sobotka: Whoops, should be `0.0722`.

sobotka:
```suggestion
	return (0.2126*c.r + 0.7152*c.g + 0.0722*c.b);
```

Issue #892, opened 2019-04-30
`makeCurrent()` call with no surface argument causes black/blank Viewers


follower:
I was wondering if the following `makeCurrent()` call with no surface argument is intentional?

https://github.com/olive-editor/olive/blob/963a20af37d78f038b69591f7879c84f744c9df2/ui/viewerwidget.cpp#L548

The docs for `makeCurrent(...)` mention:

> `bool QOpenGLContext::makeCurrent(QSurface *surface)`
> Makes the context current in the current thread, against the given surface. [...]
> **If surface is 0 this is equivalent to calling `doneCurrent()`.**
> -- <https://doc.qt.io/qt-5/qopenglcontext.html#makeCurrent>

While the docs for `doneCurrent()` mention:

> `void QOpenGLContext::doneCurrent()`
> Convenience function for calling `makeCurrent` with a 0 surface.
> **This results in no context being current in the current thread.**
> -- <https://doc.qt.io/qt-5/qopenglcontext.html#doneCurrent>

Additionally the docs for `paintGL()` mention:

> `void QOpenGLWidget::paintGL()`
> [...] **There is no need to call `makeCurrent()` because this has already been done when this function is called.**
> -- <https://doc.qt.io/qt-5/qopenglwidget.html#paintGL>

#### Impact and solution

Based on the assumption that other people have Olive running successfully, it seems like the effect of having no current OpenGL context may be undefined/system-dependent but in my case the call to `makeCurrent()` with no argument causes the "Media Viewer" and "Sequence Viewer" to not render anything other than solid black.

Removing the call to `makeCurrent()` appears to fix the issue--and takes a *very* long time to discover even if one's dealt with the "interesting" design & documentation choices of Qt's OpenGL support before. :) Which I've had the misfortune to have...

This may not be of the greatest concern because the platform it occured on is Mac OS X 10.9 with Qt 5.6 but I thought I'd at least mention it as it might cause issues elsewhere too.

This appears to be the commit where this code was introduced: https://github.com/olive-editor/olive/commit/e1ec2f6fbe26112c6cc8f7530a56d5a7da5b8a18#diff-25489ab39aa89c80d97e7872276716a7R450

#### Potential related issue

On a semi-related note, I noticed this documentation regarding `swapBuffers()` (which Olive doesn't seem to use?) and vague "To be able to support certain platforms" remark:

[...]

PR #893, opened 2019-04-30
Remove makeCurrent() call that causes no current context

follower:
"If surface is 0 this is equivalent to calling `doneCurrent()`. [...] This results in no context being current in the current thread."

For me the `makeCurrent()` call led to the "Media Viewer" and "Sequence Viewer" not rendering anything other than solid black.

See for more details: <https://github.com/olive-editor/olive/issues/892>


follower:
I...kinda forgot I'd already made this PR in the browser. :D Superseded by #898 which uses a named branch.

PR #904, opened 2019-05-03
Alpha Fixed for Crop Invert

cgvirus:


cgvirus:
response to https://github.com/olive-editor/olive/pull/880

unfa:
It is better, but not perfect:

Screenshot_20190504_165206

Note the strange edge on the cropped side - that migth however be an issue of the Chroma Key.

cgvirus:
Olive is currently broken finally. Let's wait for nodes to be processed for some days. I afraid it's the threshold of chroma key...But it has been fixed in nodes...So, no need to hack anymore.
I am closing this fix as conflict with new codes. Please save the file as back up if you need it urgently. Thanks.

unfa:
Thanks! I'm gonna abstain from installing new version of Olive right now :D

PR #926, opened 2019-05-08
Allow rendering of sequence based on tab visibility

alcomposer:


itsmattkc:
Wow is this the correct way how to check QDockWidget visibility? I could have sworn `isVisible` did this too but maybe there's a Qt bug or platform inconsistency? Will have to test this further

alcomposer:
IIRC `isVisible` just returns if the widget is actually visible. Technically tabbed widgets are visible, but placed inside a tab. If the widget were closed then `isVisible` would return 0.

This was the simplest way I researched based on the problem.

PR #938, opened 2019-05-09
Updated CZ_cs translation

mcDandy:
Its likely, that there might be still some errors.


itsmattkc:
Added to 0.1.x branch

PR #975, opened 2019-05-24
Add a new language (Chinese)

liulitchi:
Sorry for I am not familiar with Qt, nor with qmake. I  have to choose a stupid way: copy a ts file, and edit it with gedit(text editor0. It tooks two days. Not sure if it is available, if it helps, I will continue to correct this translation in the future.


BrimsonBhin:
I notice there is a handful of Chinese Olive users. Thanks for this PR

itsmattkc:
Thank you for the translation work. I believe your translation was added with #985 but correct me if I'm wrong.

PR #983, opened 2019-05-30
My ugly translation for Chinese.

kenneth-Q:
I hope this could help you.
My very very ugly translation for Simplified Chinese and Traditional Chinese.


kenneth-Q:
I found any mistake. I will fix later.
When will the latest version be released?

PR #991, opened 2019-06-05
turkish translation

usolmz:
turkish translation


itsmattkc:
Thank you for the translation work. I'm adding translations to the 0.1.x branch as Olive `master` is currently being overhauled and won't be usable for some time - this way people will be able to use the translations now rather than when the `master` is ready. Your translation files have been added to `0.1.x`.

PR #996, opened 2019-06-11
Initialize field values in frei0r-plug-ins with the default value

hzeller:
provided by the plug-in.

The plug-in might have non-zero default values (e.g. some fields are set to 100% by default), so initialize the fields we create with values from the freshly initialized plug-in to have a sensible starting point for the user.

(this addresses the most common types, double and boolean, but could be extended to Color and String if there are typically non-zero/empty values to be expected).


hzeller:
(it seems, this frei0r plugin subystem is not compiled anymore with the current refactoring state, but this patch is working fine with an older version, e.g. 00e86a4eeba1e171b97be4cfbd8ce2a5829a13a4 )

sobotka:
https://github.com/olive-editor/olive/wiki/Frei0r-Support-EOL

PR #1020, opened 2019-06-30
suggest: Event based design

orange4glace:
This PR suggests a core design of current rewriting of Olive, especially in how modularize should be done.

First, let's see how directory structure looks like. (.cpp files are omitted for convenience.)
```
.
+-- app
|   +-- platform
|      +-- theme
|         +-- themeservice.h
|         +-- themeservice-impl.h
|   +-- project
|      +-- sequence
|         +-- sequence.h
|         +-- track.h
|         +-- clip.h
|   +-- widget
|      +-- timeline
|         +-- timelinewidget.h
|         +-- sequenceview.h
|         +-- trackview.h
|         +-- clipview.h
```

If you compile it successfully, you can see,

There are three main philosophy
1. **Use event to communicate**
2. **Do not make circular dependency as possible**
3. **Use Service and Dependency Injection**

## **1. Use event to communicate**
Usually Model should not know about View. One common way to achieve it is event based design.
Qt supports a great builtin event system. *Signal and Slot*. See how `Track` utilize it.

**track.h**
```c++
class Track : public QObject {
  Q_OBJECT

private:
  std::set<Clip*> clips_;
  
  void doAddClip(Clip* const clip);
  void doRemoveClip(Clip* const clip);

public:
  Track();

  void addClip(Clip* const clip);
  void removeClip(Clip* const clip);
  bool hasClip(Clip* const clip) const;

  const std::set<Clip*>& clips();

signals:
  void onDidAddClip(Clip* const clip);
  void onWillRemoveClip(Clip* const clip);

};
```

`Track` has two events, `onDidAddClip` and `onWillRemoveClip`. Each event will be emitted when clip is added or will be removed.

**track.cpp**
```c++
void Track::doAddClip(Clip* const clip) {
  if (hasClip(clip)) return;
  clips_.insert(clip);
  onDidAddClip(clip);
}

void Track::doRemoveClip(Clip* const clip) {
  if (!hasClip(clip)) return;
  onWillRemoveClip(clip);
  clips_.erase(clip);
}
```

So everytime `Track` has some changes, the corresponding `signal` will be emitted and it will be consumed in somewhere. For example, `TrackView` can consume these events by,

**trackview.cpp**
```c++
TimelineWidgetTrackView::TimelineWidgetTrackView(
  Track* const track ...
  // Let just omit unrelated things right now.

  auto& clips = track->clips();
  for (auto& clip : clips) handleDidAddClip(clip);
  QObject::connect(track, &Track::onDidAddClip, this, &TimelineWidgetTrackView::handleDidAddClip);
  QObject::connect(track, &Track::onWillRemoveClip, this, &TimelineWidgetTrackView::handleWillRemoveClip);
}

void TimelineWidgetTrackView::handleDidAddClip(Clip* const clip) {
  // Create and add Clip view
  auto view = new TimelineWidgetClipView(this, clip, scrollView_, theme_service_);
  clipViews_.emplace(clip->id(), view);
}

void TimelineWidgetTrackView::handleWillRemoveClip(Clip* const clip) {
  // Delete Clip view
  clipViews_.erase(clip->id());
}
```

`TimelineWidgetTrackView` renders a track as a UI in **Timeline widget**. It listens `Track::onDidAddClip` and `Track::onWillClip` so when tracks are added or removed, Track UIs are also added or removed.

You can use lambda function as `slot` when needed.

Notice that `TimelineWidgetTrackView` is a **View** for `Track` used in **Timeline widget**. If you want to display any model in a UI form, you must create corresponding View to display it and every view related properties should be only there.

## **2. Do not make circular dependency as possible**
This is rather usual one. Circular dependency makes source hard to understand and increases coupling.
Whenever possible, makes them **Top to Bottom**.

For example, in Sequence model,
`Sequence --> Track --> Clip`
`Sequence` knows `Track` and `Clip`.
`Track` knows `Clip`
but `Clip` never knows `Track` and `Sequence`.
`Track` never knows `Sequence`.

## 3. **Use Service and Dependency Injection**
Notice that there's a file named `themeservice.h`
Although it's been a few days I've joined Olive discord channel, One of the request that I've seen many time was about Theme support.
One possible way to implement this is creating a `ThemeService`.
`ThemeService` provides `Theme`.

**themeservice.h**
```c++
class IThemeService : public QObject {
  Q_OBJECT

protected:
  inline IThemeService() {}

public:
  virtual void setTheme(Theme const& theme) = 0;
  virtual const Theme& getTheme() const = 0;

signals:
  void onDidChangeTheme(Theme const& theme) const;

};
```

We can get `Theme` from `ThemeService`. Sometimes theme can get changed. Notice that `ThemeService` has also an event, `onDidChangeTheme`. UIs can listen this event and when theme is changed, they can repaint themselves. For example, in `TimelineWidgetClipView`,

**clipview.cpp**
```c++
TimelineWidgetClipView::TimelineWidgetClipView(
  QWidget* const parent,
  Clip* const clip,
  SequenceScrollView* const scrollView,
  // Provide ThemeService
  IThemeService* const theme_service) : 
  QWidget(parent), clip_(clip), focused_(false), scrollView_(scrollView), theme_service_(theme_service) {
... some codes
}

... some codes

void TimelineWidgetClipView::paintEvent(QPaintEvent* event) {
  // Get current theme
  auto& theme = theme_service_->getTheme();
  QPainter p(this);
  p.fillRect(0, 0, width(), height(),
    // Draw with theme color
    focused_ ? theme.secondaryColor() : theme.primaryColor());
  
  QWidget::paintEvent(event);
}
```

One problem is "**How to provide ThemeService?**"
The simplest way is using *Singleton*. However singleton pattern is usually bad. It makes test hard and increases coupling.
Sometimes `Service locator` is used but it also increase coupling and decreases source readability.
Using DI framework may be the best solution but so far, I haven't found any good DI framework for C++. [Google fruit](https://github.com/google/fruit) is one of them.

There's no go-to solution yet. For the time being, I prefer **Use singleton only when you must need**. (The naming is little awkward so see the source below)

**core.cpp**
```c++
... some codes
  // Initialize ThemeService singleton
  olive::ThemeService::Initialize();
... some codes
```

**timeline.cpp** (pannel/project/timeline)
```c++
TimelinePanel::TimelinePanel(QWidget *parent) :
  PanelWidget(parent)
{
  ... some codes
  auto widget = new olive::TimelineWidget(this, olive::ThemeService::instance());
  setWidget(widget);
  ... some codes
}
```
Notice that `TimelineWidget` depends on `ThemeService` and `ThemeService` instance is being injected. The point is `ThemeService` instance itself comes from `ThemeService::instance()`, which is singleton instance.
But once singleton instance is injected to `TimelineWidget`, `TimelineWidget` can pass `ThemeService` instance without using singleton. For example,

**timelinewidget.cpp**
```cpp
TimelineWidget::TimelineWidget(
  QWidget* parent,
  IThemeService* const theme_service) :
  QWidget(parent),
  sequence_(nullptr),
  sequence_view_(nullptr),
  sequence_scroll_view_(nullptr),
  // Store ThemeService instance as private member
  theme_service_(theme_service) {
}

... some codes

// setSequence creates a `TimelineWidgetSequenceView` which depends on `ThemeService`
void TimelineWidget::setSequence(Sequence* sequence) {
  if (sequence_view_ != nullptr) {
    sequence_ = nullptr;
    delete sequence_view_;
    delete sequence_scroll_view_;
  }
  if (sequence == nullptr) return;
  sequence_ = sequence_;
  sequence_scroll_view_ = new SequenceScrollView(this, sequence);
  // ThemeService instance can be injected since TimelineWidget has theme_service_ as a private member.
  sequence_view_ = new TimelineWidgetSequenceView(sequence_scroll_view_, sequence, sequence_scroll_view_, theme_service_);
}
```

Like `ThemeService`, a specific functions can be provided by using Service pattern. For example, `FileService` can be implemented to manage file system or `TimelineWidgetService` can be implemented to manage which timeline widget is currently focused etc..

### Some small improvements may be considered

1. **namespace guard**
`namespace olive { ...`

2. **[include guard prefix](https://google.github.io/styleguide/cppguide.html#The__define_Guard)**
It is rarely possible but include guard might collide with other libraries
`#ifndef OLIVE_~~~_H_`

Please see the source code for more information.

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -32,6 +32,7 @@ add_subdirectory(tool)
 add_subdirectory(ui)
 add_subdirectory(widget)
 add_subdirectory(window)
+add_subdirectory(platform)
 
 set(OLIVE_TARGET "olive-editor")
 if(APPLE)
--- a/app/core.cpp
+++ b/app/core.cpp
@@ -36,6 +36,8 @@
 #include "ui/style/style.h"
 #include "widget/menu/menushared.h"
 
+#include "platform/theme/themeservice-impl.h"
+
 Core olive::core;
 
 Core::Core() :
@@ -73,6 +75,9 @@ void Core::Start()
   if (!args.isEmpty()) {
     startup_project_ = args.first();
   }
+
+  // Initialize Services
+  olive::ThemeService::Initialize();
 
 
 
--- a/app/panel/timeline/timeline.cpp
+++ b/app/panel/timeline/timeline.cpp
@@ -20,9 +20,30 @@
 
 #include "timeline.h"
 
+#include "platform/theme/themeservice-impl.h"
+#include "widget/timeline/timelinewidget.h"
+#include "project/sequence/sequence.h"
+#include "project/sequence/track.h"
+#include "project/sequence/clip.h"
+
+#include <iostream>
+
 TimelinePanel::TimelinePanel(QWidget *parent) :
   PanelWidget(parent)
 {
+  // Create Sequence mock data
+  auto sequence = new olive::Sequence();
+  sequence->addTrack();
+  auto track1 = sequence->addTrack();
+  auto track2 = sequence->addTrack();
+  track1->addClip(new olive::Clip(30, 150));
+  track2->addClip(new olive::Clip(120, 260));
+
+  auto widget = new olive::TimelineWidget(this, olive::ThemeService::instance());
+  widget->setSequence(sequence);
+
+  setWidget(widget);
+
   Retranslate();
 }
 
--- a/app/platform/CMakeLists.txt
+++ b/app/platform/CMakeLists.txt
@@ -1,0 +1,6 @@
+add_subdirectory(theme)
+
+set(OLIVE_SOURCES
+  ${OLIVE_SOURCES}
+  PARENT_SCOPE
+)
--- a/app/platform/theme/CMakeLists.txt
+++ b/app/platform/theme/CMakeLists.txt
@@ -1,0 +1,8 @@
+set(OLIVE_SOURCES
+  ${OLIVE_SOURCES}
+  platform/theme/theme.h
+  platform/theme/themeservice.h
+  platform/theme/themeservice-impl.h
+  platform/theme/themeservice-impl.cpp
+  PARENT_SCOPE
+)
--- a/app/platform/theme/theme.h
+++ b/app/platform/theme/theme.h
@@ -1,0 +1,63 @@
+#ifndef _OLIVE_THEME_H_
+#define _OLIVE_THEME_H_
+
+#include <string>
+#include <QColor>
+
+using namespace std;
+
+namespace olive {
+
+class Theme {
+
+private:
+  QColor primaryColor_;
+  QColor secondaryColor_;
+  QColor surfaceColor_;
+  QColor surfaceDarkColor_;
+  QColor surfaceBrightColor_;
+  QColor surfaceTextColor_;
+  QColor surfaceContrastColor_;
+
+
+public:
+  Theme(
+    QColor primaryColor,
+    QColor secondaryColor,
+    QColor surfaceColor,
+    QColor surfaceDarkColor,
+    QColor surfaceBrightColor,
+    QColor surfaceTextColor,
+    QColor surfaceContrastColor) : 
+    primaryColor_(primaryColor),
+    secondaryColor_(secondaryColor),
+    surfaceColor_(surfaceColor),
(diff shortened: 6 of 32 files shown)


itsmattkc:
Thank you for putting the time into this demonstration. Indeed I agree with your code design and think for the most part I've been heading in that direction in the current rewrite (with the exception of a handful of ideas that would probably be good to implement).

I won't be able to merge this (I'm not sure if this was intended to be?) since the internal timeline design will actually be a very different to how it was written before/here. But this PR does serve as a good reminder of how to keep good practices when writing it up.

PR #1025, opened 2019-07-02
Icons Redesign

Fatih20:
Here's the new pull request since the previous one was automatically closed after the merge between rewrite and master. Only file "bold.svg" hasn't been changed since it can't be opened with inkscape. @itsmattkc has said some criticism about this redesign and I would like to hear specifically the changes that is needed.


Simran-B:
I'm not a fan of the rounded style, because it makes some of the icons really hard to recognize. Another general point to discuss is whether icons should be monochrome, duotone or something else.

Fatih20:
I think it should be duotone because there are certain shapes that are on top of each other.

Simran-B:
It doesn't necessarily require two colors to stack objects without melting them together visually.
Many material design icons work as filled, outline and two-tone variant: https://material.io/tools/icons/

Fatih20:
> It doesn't necessarily require two colors to stack objects without melting them together visually.
> Many material design icons work as filled, outline and two-tone variant: https://material.io/tools/icons/

The icons you linked are used in mobile devices and web. They are displayed in bigger size than what we use for our use case, which is icon for tool. I could try using outline or gaps but they wouldn't be too visible considering that the icons is pretty small.

Fatih20:
https://drive.google.com/open?id=1PaGfC8kclkOfPLMO9IXGv-dbdTOr89Pj Here's an updated icon set. Just need to make new commit for it.

PR #1069, opened 2019-08-24
Replacing Turkish translation file to original version

tecnotercio:
The "olive_tr.ts" file released in version 0.1.1 is actually the HTML code of a GitHub page.
The version available here is the same as pull request #991.


itsmattkc:
Thank you, I must have made a mistake when merging the files.

PR #1101, opened 2019-09-29
Fix linkage error

sobotka:
Linkage error without empty definition.

```
[100%] Linking CXX executable olive-editor
/usr/bin/ld: CMakeFiles/olive-editor.dir/project/item/footage/imagestream.cpp.o: in function `ImageStream::set_colorspace(QString const&)':
imagestream.cpp:(.text+0x31b): undefined reference to `ImageStream::ColorSpaceChanged()'
collect2: error: ld returned 1 exit status
```

### DIFF
--- a/app/project/item/footage/imagestream.cpp
+++ b/app/project/item/footage/imagestream.cpp
@@ -84,3 +84,8 @@ void ImageStream::ColorConfigChangedSlot()
   // FIXME: Update colorspace correctly
   colorspace_.clear();
 }
+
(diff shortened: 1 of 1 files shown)


itsmattkc:
Ahh I forgot the QObject derivation on the Stream class (Qt MOC is supposed to define this function automatically). I'll be able to push the fix soon.

itsmattkc:
45bfad2b2002d866f763fe0c7379d99f1965b7c4 should fix this

PR #1111, opened 2019-10-26
Code Cleanup: common: removing default constructors

lightyear15:
As C++11 is the adopted standard, trivial constructors are not needed,
compiler can generate it for us.

### DIFF
--- a/app/common/rational.h
+++ b/app/common/rational.h
@@ -51,10 +51,7 @@ public:
       numer = intType(0);
   }
 
-  rational(const rational &rhs)
-    :numer(rhs.numer), denom(rhs.denom)
-  {
-  }
+  rational(const rational &rhs) = default;
 
   rational(const AVRational& r);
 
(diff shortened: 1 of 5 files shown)


sobotka:
Nice work!

@itsmattkc, what do you think?

lightyear15:
> Nice work!
> 
> @itsmattkc, what do you think?

That's unfortunate, I thought things in this project run a bit faster.

sobotka:
Currently one developer. Mitigating burnout.

lightyear15:
> Currently one developer. Mitigating burnout.

Alright I see. When you have time then we might discuss some ideas I have, like for example rearranging the cmake file so that we can add some unit tests.
I have some ideas on refactoring classes like ``rational.cpp`` I would like discuss as well.
Cheers.

sobotka:
If you provide the branch, I’d think Matt would certainly consider it.

I know he’s spoken about unit tests before, including colour transforms etc., so the effort would be more than likely welcome. I’d say go for it.

sobotka:
Grab that 1 day valid link for Slack @lightyear15

itsmattkc:
@lightyear15 I'm sorry, I've been meaning to get around to this but have been a little preoccupied lately this past week. It all seems good, so I have no problem merging in.

PR #1112, opened 2019-11-02
Fixed not working on AMD GPUs

unfa:
Fixed implicit integer to float conversion which is rejected on AMD cards (because Nvidia allows sloppy code, that's not compliant to the GLSL spec).

See "Nvidia and Types" section here:
https://www.khronos.org/opengl/wiki/GLSL_:_common_mistakes


unfa:
Thanks!

PR #1114, opened 2019-11-08
Task and TaskThread small improvements

lightyear15:
just small stuff to improve readability


itsmattkc:
The Task class has seen fairly significant rewrites recently. Feel free to check it now for improvements if you'd like.

PR #1117, opened 2019-11-15
rename language files so that Fedora macro %find_lang can handle it

martinkg:
Hi,

on Fedora, the macro % find_lang will not find all the language files, if they are in the format olive_pt-BR.ts. That's why I have renamed the files to olive_pt_BR.ts.
Can you please check my patch and commit it.

Regards
Martin


tecnotercio:
@martinkg, the Qt library requires the language and country to be separated by a hyphen rather than an underscore:

https://wiki.qt.io/Basics_of_Locales

Olive does not recognize translation files renamed with underscore.

kkofler:
That is not an inherent limitation of the Qt translation system, but of your application.

E.g., Calamares uses underscores as language-country separators in the file names: https://github.com/calamares/calamares/tree/master/lang and they load just fine.

Since everything else on *nix (all the gettext stuff) uses that underscore naming, I would recommend fixing your application to use that naming, too.

kkofler:
The issue is actually the `lastIndexOf` at:
https://github.com/olive-editor/olive/blob/52d03aacd68564974e42ba7667e58bb04ce50d15/dialogs/preferencesdialog.cpp#L528

That splits the name into, e.g., `olive_pt` and `BR` instead of `olive` and `pt_BR`. Replace that `lastIndexOf` with `indexOf` and it should detect the files with the undercore locale naming just fine.

itsmattkc:
Hi, sorry for such a late response. I've taken note of this and the languages will use this naming scheme in the next release soon.

PR #1118, opened 2019-11-21
Add batch exporting feature (with default settings)

morrolinux:
This would allow any user to directly export a project from commandline invocation (`option -e`)
Particularly useful for scripting/batch export


morrolinux:
While being extremely useful for me as of yet, it would also be nice to be able to export headlessly (without the need for the interface to open at all) for thinks like distributed export on Amazon AWS. Do you think it will be possibile to have headless export on the new version of Olive, @itsmattkc ?

sobotka:
Headless rendering is indeed part of the vision.

PR #1125, opened 2020-02-21
Update README.md

alexmyczko:


itsmattkc:
I'm having trouble determining what this link does, we already offer binaries at https://github.com/olive-editor/olive/releases

BrimsonBhin:
I think it is for getting releases outside of the Appimage and exe.. Though a lot of those 3rd party repos are dead

alexmyczko:
I am the maintainer of the package for Debian, Canonical/Ubuntu, Raspbian, Devuan take copies from Debian, Mint uses Ubuntu repo. All these install software with a command like:
apt install softwarename

itsmattkc:
Indeed we don't, but the wording "Getting Binaries" makes it sound like it's the only/main place to get binaries and that they are officially maintained by us, neither of which are true.

I'd be more inclined to add this link to the "unofficial" section of [this page](https://olivevideoeditor.org/linux.php).

PR #1126, opened 2020-02-28
Support to x86_32?

TheNeedJedi:
Hello! I work with olive and he is my default video editor but I can't use editor updates because my pc is 32. Can I bring support for x32? Thanks


itsmattkc:
The continuous build is not ready for use. 0.1.x (the latest "stable") is still available for 32-bit.

PR #1127, opened 2020-03-01
Insert into timeline

lenzwerf:
There should be a shortcut for inserting a part of a clip into the timeline. Now you have to drag the clip, selected by in/out points in the media viewer, to the timeline. Every edit program has a shortcut for this. Normally 'V' for insert in the timeline and 'B' for overwrite.
Serious editors use in and out points and shortcuts. Dragging from the source monitor or whatever is slow and causes RSI.


prokoudine:
V is already taken though

itsmattkc:
This section is not for feature requests. Please read the README for more information. You may make feature requests when the issues tab re-opens.

PR #1132, opened 2020-03-27
Update timelinecoordinate.cpp

destinyx-code:


app/timeline/timelinecoordinate.cpp
```diff
@@ -3,7 +3,7 @@
   Olive - Non-Linear Video Editor
   Copyright (C) 2019 Olive Team
 
-  This program is free software: you can redistribute it and/or modify
```
v0idzz: It's free as in freedom. No need for that 'and open source' I believe.

app/timeline/timelinecoordinate.cpp
```diff
@@ -3,7 +3,7 @@
...
+  This program is free and open source software: you can redistribute it and/or modify
```
v0idzz: It's free as in freedom. No need for that 'and open source' I believe.

sobotka:
Free software has a unique definition. “Open source” has a different definition.

PR #1133, opened 2020-04-03
Changing font family doesn't work

cyump07:
Hi @itsmattkc, nice video editor.
I'm new to qt and tried to make some changes to the Olive video editor, and after the build I noticed that I'm not able to change the font family In the rich text editor. The font families are all there but nothing happens after I select a font and click OK. When I go back to edit the title the default font family is selected. 

fonts

All other functions work great.. I've tried this on all the releases and built on Win and Mac.
Am I doing something wrong, or what could be the issue here?


manokara:
What exactly did you change? That's a massive pull request. Olive is currently undergoing a rewrite, which is why _Issues_ are disabled for now and people are not supposed to file bug reports yet.

cyump07:
Sorry, haven't really worked on github before, just tried to report a bug, and here's where the site brought me. You can close this if it's in the wrong place, but if you could point me in the right direction would be great.

sobotka:
This isn't an issue tracker and Olive 0.1.x is not being developed. There currently are no trackers open.

manokara:
It's not only the wrong place, but also the wrong time. Like I said, people are not supposed to be reporting bugs or hacking into the code at this moment, which is why the right place (`Issues` tab) isn't available. If you're trying to learn Qt through an existing codebase, I would recommend something more stable like [Okteta](https://github.com/KDE/okteta).

I'm not one of the maintainers, so you'll have to close it yourself. There should be a `Close pull request` button bellow the comment box.

PR #1137, opened 2020-04-22
Upgrade shaders from GLSL 110 to 150 (OpenGL 3.2)

Simran-B:
Useful information:
- https://github.com/mattdesl/lwjgl-basics/wiki/GLSL-Versions
- http://www.shaderific.com/blog/2014/3/13/tutorial-how-to-update-a-shader-for-opengl-es-30
- http://retokoradi.com/2014/03/30/opengl-transition-to-core-profile/

Summary:
- `#version 110` -> `#version 150`
- `attribute` -> `in` in vertex shaders
- `varying` -> `out` in vertex shaders
- `varying` -> `in` in fragment shaders
- `gl_FragColor` -> `fragColor` in fragment shaders, and a declaration at the top `out vec4 fragColor;`
- `texture2D()` -> `texture()`
- Formatting

Compiled successfully on my local machine. Footage and sequence viewer were working as before in a brief test run, so I think I didn't break anything.

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -47,7 +47,7 @@ int main(int argc, char *argv[]) {
   QSurfaceFormat format;
   format.setVersion(3, 2);
   format.setDepthBufferSize(24);
-  format.setProfile(QSurfaceFormat::CompatibilityProfile);
+  format.setProfile(QSurfaceFormat::CoreProfile);
   QSurfaceFormat::setDefaultFormat(format);
 
   // Try to share OpenGL contexts
--- a/app/node/math/math/math.cpp
+++ b/app/node/math/math/math.cpp
@@ -133,7 +133,7 @@ QString MathNode::ShaderFragmentCode(const NodeValueDatabase &input) const
     NodeParam* tex_in = (type_a == NodeParam::kTexture) ? param_a_in_ : param_b_in_;
     NodeParam* mat_in = (type_a == NodeParam::kTexture) ? param_b_in_ : param_a_in_;
 
-    operation = QStringLiteral("texture2D(%1, (vec4(ove_texcoord, 0.0, 1.0) * %2).xy)").arg(tex_in->id(), mat_in->id());
+    operation = QStringLiteral("texture(%1, (vec4(ove_texcoord, 0.0, 1.0) * %2).xy)").arg(tex_in->id(), mat_in->id());
 
   } else {
     switch (GetOperation()) {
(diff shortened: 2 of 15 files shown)


Simran-B:
@itsmattkc is there anything in OpenGL 3.3 that would be useful in Olive?
http://rastergrid.com/blog/2010/03/a-brief-preview-of-the-new-features-introduced-by-opengl-3-3-and-4-0/
esp. ARB_explicit_attrib_location, ARB_blend_func_extended and ARB_timer_query.

BrimsonBhin:
Olive could benefit from stuff such as bit-wise operators for integers in some shaders. Performance benefit depends on what you're coding

Simran-B:
Which shaders would that be? They are all float based aren't they? And how would you utilize bit-wise operations?

BrimsonBhin:
> Which shaders would that be? They are all float based aren't they?

I would say the blur shaders, since it involves looping a texture's `pixel_coords`. MattKC made it 2 iterations to save the loops the GPU has to do. 64+64 vs 64*64.

I am not very experienced with bit-wise operations, but this user used such operations to reduce the loops? https://www.shadertoy.com/view/ltScRG.

The user also used `textureLod()`, which is [only supported from GLSL 1.30 onward](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/textureLod.xhtml) :)

BrimsonBhin:
> Which shaders would that be? They are all float based aren't they?
> And how would you utilize bit-wise operations?

I am guessing the blur shaders. I am not experienced with bit-wise operations yet. I am thinking about porting a [faster, single-pass Gaussian blur from ShaderToy that uses bit-wise operations for loops](https://www.shadertoy.com/view/ltScRG#). Based on the Shadertoy's comments, the author used operations to reduce the loop. This could be beneficial for people to test different GLSL/OpenGL features on Olive. The ShaderToy also uses mip-maps via `textureLod`, which is also only supported by GLSL 1.30 onward :)

As with Olive's blur shader, it samples a whole texture and loops a texture's `pixel_coords`. Olive executes the blur in 2 iterations to reduce the amount of loops the GPU has to do (64*64 vs 64+64). It is unnecessary to make the current blur shader use bit-wise operations because the its loop already uses float values optimized for linear sampling.

Simran-B:
@BrimsonBhin I don't see any bitwise operations in that shader? How are bitwise operations on IEEE 754 floating point representations even defined? In C++ it's not allowed and:

> there's [...] no meaningful expression in C++ requiring bitwise ops on floats

https://stackoverflow.com/questions/1723575/how-to-perform-a-bitwise-operation-on-floating-point-numbers

Looking at the latest GLSL spec, they are only defined for integers:

> The bitwise operators and (&), exclusive-or (^), and inclusive-or (|). The operands must be of
type signed or unsigned integers or integer vectors. 

https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.60.pdf

Bitwise operations on floats don't make any sense to me, because the binary representation is comprised of 3 components. How could treating them like a single thing result in anything meaningful?

I found the following Python code, which seems to implement bitwise floating point operations. Not sure what it tries to accomplish and how it behaves. But it seems to treat sign (bit), mantissa and exponent separately in some of the functions. It's an awful lot of code however, so I don't think that there is anything to win in terms of performance:

http://code.activestate.com/recipes/577967-floating-point-bitwise-operations/

Regarding blur shaders: using mip mapping (esp. for large blur kernels) is an option to consider, but I have doubts about the quality. The shader you linked and the original implementation produce different results. If the results look different, then the one using the mip mapping optimization would be an incorrect Gaussian blur if you will. In practice it may or may not matter to you, but it's something the user should be able to control. Or is it just a matter of adjusting sigma to take mip-mapping into account?

I also noted that the original shader description says:

> Much heavier than two-pass, but useful when you can't do two passes.

https://www.shadertoy.com/view/4tSyzy

Not sure whether the optimized one-pass shader has the same complexity and practical performance as the best two-pass implementation or not. But I guess two-pass is the better choice, and Olive has no requirement for a one-pass shader or does it?

PR #1140, opened 2020-04-29
MSVC: Build with Multiple Processes (CMAKE_CXX_FLAGS /MP)

Simran-B:
Enables parallel compilation under Windows for MSVC compiler. `/MP` will spawn one process per effective processor (virtual CPU core count), so `%NUMBER_OF_PROCESSORS%`. `/MP4` would use 4 processes etc.

https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=vs-2017

Note to self: Docker for Windows with Hyper-V isolation uses 2 virtual cores by default. Can be increased like `docker run -it --cpu-count %NUMBER_OF_PROCESSORS% olive`.

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -92,6 +92,7 @@ if(MSVC)
     /external:anglebrackets
     /external:W0
     "$<$<CONFIG:RELEASE>:/O2>"
+    "$<$<COMPILE_LANGUAGE:CXX>:/MP>"
   )
 else()
   target_compile_options(


eszlari:
Doesn't the Windows build use `-GNinja`?

Simran-B:
Well, I don't. I use a Docker image and MSVC

```
cmake .. -G "Visual Studio 15 2017 Win64 -DOCIO_LOCATION=C:\Tools\vcpkg\packages\opencolorio_x64-windows -DOIIO_LOCATION=C:\Tools\vcpkg\packages\openimageio_x64-windows -DOPENEXR_LOCATION=C:\Tools\vcpkg\packages\openexr_x64-windows -DFFMPEG_ROOT=C:\TEMP\ffmpeg-4.2.2-win64-dev\ffmpeg-4.2.2-win64-dev"
cmake --build . --config RelWithDebInfo
```

The AppVeyor CI uses NMake and jom:
https://github.com/olive-editor/olive/blob/master/.appveyor/build.bat#L37

Anyway, the flag is added for MSVC only, so it won't affect any other compilers, but tremendously speedup compilation if you use Visual Studio.

eszlari:
> Well, I don't. I use a Docker image and MSVC
> 
> ```
> cmake .. -G "Visual Studio 15 2017 Win64" -DOCIO_LOCATION=C:\Tools\vcpkg\packages\opencolorio_x64-windows -DOIIO_LOCATION=C:\Tools\vcpkg\packages\openimageio_x64-windows -DOPENEXR_LOCATION=C:\Tools\vcpkg\packages\openexr_x64-windows -DFFMPEG_ROOT=C:\TEMP\ffmpeg-4.2.2-win64-dev\ffmpeg-4.2.2-win64-dev
> cmake --build . --config RelWithDebInfo
> ```

AFAIK that uses MSBuild as build tool.

> The AppVeyor CI uses NMake and jom:
> https://github.com/olive-editor/olive/blob/master/.appveyor/build.bat#L37

To be precise: it outputs NMake files and then compiles with jom (which is a multithreaded NMake replacement).

Between ninja, MSBuild & jom, ninja should be the fastest. It's also avaible on [appveyor](https://www.appveyor.com/docs/windows-images-software/) and there are binaries on [Github](https://github.com/ninja-build/ninja/releases).

Simran-B:
Yeah, I think it uses MSBuild, which then spawns cl.exe (Microsoft C-C++ Compiler Driver), which I tend to call MSVC (as in Microsoft's Visual C++ compiler). I assume that `if(MSVC)` in CMake only has an effect if Visual Studio is used as generator, but let me know if that's not the case.

There's no reason for me to switch to ninja, as I want to be able to work in Visual Studio, e.g. for profiling. Might be an option for the CI, but honestly, what we have at the moment works and compilation time is not a concern. Anyway, I don't see anything speaking against merging this PR, which would lift the need to patch the CMakeLists.txt locally to enable parallel compilation if MSVC is used.

eszlari:
> I assume that `if(MSVC)` in CMake only has an effect if Visual Studio is used as generator, but let me know if that's not the case.

No: https://cmake.org/cmake/help/latest/variable/MSVC.html

> There's no reason for me to switch to ninja, as I want to be able to work in Visual Studio, e.g. for profiling.

VS supports cmake natively since VS2017.

> compilation time is not a concern

For a "edit compile debug cycle", the fact should come in handy, that ninja has AFAIK the fastest incremental builds of all build systems.

sobotka:
@eszlari Join this slack. https://join.slack.com/t/olive-editor/shared_invite/zt-dxytrx91-WLEvLaRVDNF40vtvSbAx~w

Simran-B:
Oh well, thanks for the heads up! Still not sure if adding the `/MP` flag ultimately affects Ninja or not. If Ninja spawns a compile process for each files individually, then it should not do any harm:
https://docs.microsoft.com/en-us/cpp/build/reference/mp-build-with-multiple-processes?view=vs-2017#number-of-processes
If it supplies multiple file per compiler invocation, then I'm not sure.

> VS supports cmake natively since VS2017.

Good hint, will read up on that and will probably come in handy for locale development. Can't tell if it is usable in a Docker container yet. Current CMake didn't work, because of incomplete UTF-8 support (and the codepage in a Hyper-V sandbox is hardcoded to 65001), which caused errors with the CMake copy command. Only a very recent RC version of CMake fixed the problem. If VS uses its own implementation then it's probably not affected.

> For a "edit compile debug cycle", the fact should come in handy, that ninja has AFAIK the fastest incremental builds of all build systems.

Of course. I'm not disagreeing at all, I'm just saying that for the our AppVeyor builds it will not make much of a difference. Not sure if it's using any build cache in the current configuration, but it looks like it does not. Furthermore, compilation is just one of the steps. Is does a lot more stuff that takes time, e.g. NSIS packaging. For local development, it's a whole different story and I'm happy to try out Ninja!

itsmattkc:
@eszlari actually makes a good point, I'm not sure why the AppVeyor wasn't using Ninja before, but it does now.

PR #1141, opened 2020-04-30
black screen 

Xag8032:
i am using olive for a vidio edit and i get black when i try to wach it


itsmattkc:
I'm afraid we don't take issue reports for 0.1.x anymore, but a black screen 99% of the time means your graphics card drivers are not installed/up to date.

PR #1142, opened 2020-05-02
Docs: Start working on Sphinx documentation system

Simran-B:
This is **WIP**, do not merge yet.


Simran-B:
Hm, how useful is the Doxygen API documentation at all for developers? It takes pretty long to generate and the result takes up quite a bit of storage space. For documentation build previews, only Netlify and RTD seem to viable because of free storage and bandwidth limits on e.g. GitHub Pages. Would it make sense to use some static site generator for the website + docs, and to use Sphinx only to render Doxygen XML more nicely?

ThomasWilshaw:
I've never used it, you get all you need through the IDE usually anyway. I think it will be more useful as and when it's more complete/not just function definitions.

I was thinking of adding some documentation to  header files. Should I pull request of off this branch or off master if I do?

Simran-B:
Good point, IntelliSense renders it quite useless. If we don't need Doxygen for C++/Python documentation anytime soon, then we could as well go for Hugo for hand-written documentation and the website.

This PR only adds Sphinx as documentation system. If you want to document things in header files then base it on current master.

Simran-B:
Cleaned up and pushed my local work for posterity. It's just a prototype that we won't use, but it might still be useful as a reference should we decide to use Sphinx for code documentation (essentially as Doxygen renderer) at some point.

PR #1144, opened 2020-05-03
Channel Swizzle For Waveform

ThomasWilshaw:
Adds channel swizzling (RGB and Luma) for the waveform. Possibly not the best way to do it and probably needs cleaning up a bit. I just wanted to check I was going in the right direction.

### DIFF
--- a/app/panel/scope/scope.cpp
+++ b/app/panel/scope/scope.cpp
@@ -21,6 +21,8 @@
 #include "scope.h"
 
 #include <QVBoxLayout>
+#include <QCheckBox>
+#include <QGuiApplication>
 
 #include "panel/viewer/viewer.h"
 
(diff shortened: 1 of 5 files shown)


ThomasWilshaw:
Build failure is due to `UseNativeWindowsStyling` having not been fully updated in master.

ThomasWilshaw:
Slightly messed up some commits and had to do some history editing. I think it went okay but if theres any issues that will be why.

ThomasWilshaw:
Example showing just Luma and blue channel:
image

ThomasWilshaw:
I think this is mostly done now if someone wants to take a look.

ThomasWilshaw:
Squashed to a single commit with a sensible message.

ThomasWilshaw:
I suspect this will need to be updated if you add @sobotka's changes, but I'll do that as and when that's commited.

PR #1146, opened 2020-05-04
OCIO Fixes

sobotka:
Crimps down the 3D LUT range and adds an offset. This should fix the edge sampling errors in theory.

### DIFF
--- a/app/render/ocioconf/config.ocio
+++ b/app/render/ocioconf/config.ocio
@@ -69,7 +69,7 @@ colorspaces:
       ITU BT.709 primaries based scene referred linear space.
     isdata: false
     allocation: lg2
-    allocationvars: [-12.473931188, 12.526068812]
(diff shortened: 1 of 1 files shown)


itsmattkc:
Should this fix the OCIO issue in the GPU path? I still get 0.98 on whites

sobotka:
@itsmattkc It should be _slightly better_ in the waveform, not perfect. The only way to be "perfect" would be to skip the reference space and loop directly from source to destination I believe, which is meaningless in Olive in terms of practicality and otherwise.

I intend to create a default 2.2 sRGB Display class, as you noted, it makes sense and is in fact a *majority* of display classes.

itsmattkc:
Gotcha, seems good at least for the time being.

sobotka:
When I'm involved, we need to have a _very_ low bar for "good". There's plenty to sort out configuration-wise, and this configuration should be explicitly considered a "Developer Only" configuration.

PR #1147, opened 2020-05-04
Tabbify Project Properties

ThomasWilshaw:
As per @sobotka's suggestion I've tabbified the Project Properties. Are they any design guidlines that should be followed?
image


itsmattkc:
I think I'd be inclined to just use QWidget instead of QFrame, but otherwise seems sane.

ThomasWilshaw:
Will do. For the sake of consistency do you want the same done to Preferences?

itsmattkc:
Tabs? No, it already used to be tabs, but we moved away from them because they become cumbersome with a lot of pages.

ThomasWilshaw:
Changed QFrames to QWidget and squashed commits. Should be good to go.

itsmattkc:
Merged in with 3da9643da2c142c6535dfb2cfbf4e68cc76aeeea (not sure why it didn't pick it up automatically)

Issue #84, opened 2018-06-21
Titler/on screen text


itsmattkc:

maingi2:
For some reason, .txt files are rendered onto the screen and can be used as a makeshift titler

itsmattkc:
Basic titling added, I'll create new issues for later features

PR #1148, opened 2020-05-05
Implement alt+click to reset to default value in SliderBase Issue #84

ThomasWilshaw:
I've had ago at implementing this in a fairly simple way. I've added a default value that gets set in the constructor. If you Alt click a slider the value gets set to the default value. It doesn't work on sliders where the default is not the default set in SliderBase so I need to sort that. I just wanted to check I was going in the right direction.

Fixes sobotka/Olive#84

### DIFF
--- a/app/dialog/export/codec/h264section.cpp
+++ b/app/dialog/export/codec/h264section.cpp
@@ -121,6 +121,7 @@ H264CRFSection::H264CRFSection(QWidget *parent) :
   crf_input->SetMinimum(kMinimumCRF);
   crf_input->SetMaximum(kMaximumCRF);
   crf_input->SetValue(kDefaultCRF);
+  crf_input->SetDefaultValue(kDefaultCRF);
   layout->addWidget(crf_input);
 
   connect(crf_slider_, &QSlider::valueChanged, crf_input, &IntegerSlider::SetValue);
--- a/app/dialog/export/export.cpp
+++ b/app/dialog/export/export.cpp
@@ -204,7 +204,9 @@ ExportDialog::ExportDialog(ViewerOutput *viewer_node, QWidget *parent) :
   // Set defaults
   format_combobox_->setCurrentIndex(kFormatMPEG4);
   video_tab_->width_slider()->SetValue(viewer_node_->video_params().width());
+  video_tab_->width_slider()->SetDefaultValue(viewer_node_->video_params().width());
   video_tab_->height_slider()->SetValue(viewer_node_->video_params().height());
+  video_tab_->height_slider()->SetDefaultValue(viewer_node_->video_params().height());
   video_tab_->set_frame_rate(viewer_node_->video_params().time_base().flipped());
   audio_tab_->set_sample_rate(viewer_node_->audio_params().sample_rate());
   audio_tab_->set_channel_layout(viewer_node_->audio_params().channel_layout());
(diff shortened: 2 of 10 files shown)


itsmattkc:
This is essentially what I was looking to achieve, but I would do the `Qt::AltModifier` check in SliderLabel and emit a different signal like `emit ResetDefault()` if so (plus disable dragging).

ThomasWilshaw:
There are two types of sliders, those that use the default value set in the SliderBase constructor (e.g. Position in the Orthographic Matrix node) and those that set a seperate default (e.g. Scale in the Orthographic Matrix node). This is causing issues; how do we know when to update the default value? The simplest way of fixing this I can think of is to find all the sliders that need an custom default set and explicitly call `SetDefaultValue()` with that value. Does that sound sensible? Or am I missing something?

itsmattkc:
I think you're thinking of the NodeInput constructor, not the SliderBase constructor. The NodeInput default value doesn't change, therefore the NodeParamViewWidgetBridge is the class that would call SetDefaultValue() on the slider on init.

ThomasWilshaw:
So in the NodeParamViewWidgetBridge CreateWidgets it woul be something like:

```
slider->SetDefault(input_->get_standard_value());
```

ThomasWilshaw:
That seems to have worked. I think it's just a cas eof now finding all the places where sliders are used and settign there defaults.

itsmattkc:
It would be `get_standard_value()` necessarily, I don't think there's actually access to the default values (one needs to be written in).

But yes, that's pretty much it. I'd say it's not even necessary to look for other sliders, those are the ones that matter the most.

ThomasWilshaw:
Yes I had to use `get_split_standard_values()` for the multiple float sliders. There's a few like the colour picker and some of the settings in the export dialoge that need doing as they go to odd values when you Alt click on them. For example the dimensions of the exported video default to 1 when clicked, as opposed to the project dimensions (1920x1080 or whatever).

itsmattkc:
If I were you, I'd do something like a `QVariant::isNull()` or a `bool` check to see if any default value has actually been set, and not bother setting to default if not since not all sliders are going to need a default value.

ThomasWilshaw:
Sliders requiring defaults:
 -  ~~Nodes~~
 -  ~~Timeline~~
 -  Colour Picker
 -  Export Dialog Dimensions
 - ...

ThomasWilshaw:
> 
> 
> If I were you, I'd do something like a `QVariant::isNull()` or a `bool` check to see if any default value has actually been set, and not bother setting to default if not since not all sliders are going to need a default value.

I'll give it  a go.

sobotka:
Can you edit the subject to be more clear and add a link to the relevant issue? I would love to chase this but have no clue.

ThomasWilshaw:
Linked PR to relevant issue.

itsmattkc:
Out of those, I think the following _shouldn't_ have defaults:
- Timeline
- Color Picker
- Target Bit Rate
- Target File Size
- Any in Preferences (would probably be better to have a dedicated "reset to defaults" button somewhere)

Everything else is fine, and no others come to mind that need defaults.

ThomasWilshaw:
Here's the final list of defaults:

 -  ~Nodes~ (What it's initialy set too or 0 if not)
 -  ~Export Dialog Dimensions~ (Defaults to Sequence width/height)
 -  ~Export Compression Settings~
           - ~Advanced (0)~
 - ~Speed/Duration~
        - ~Speed (100%)~
        -  ~Duration (clip length)~
 - Stream Properties Image sequence
       - ~Start Index~ (1)
       - ~End Index~ (sequence length)

All others ignore an Alt Click.

Needs cleaning up.

ThomasWilshaw:
I've noticed the new gizmos don't update when we reset to the default value but I'm not sure how to fix that yet.

itsmattkc:
I can look into gizmos. Is this ready to review and merge otherwise?

ThomasWilshaw:
Thanks I not had a chance to get my head round them yet. Yes this is ready for review.

PR #1150, opened 2020-05-05
Setlocale Windows build fix

ThomasWilshaw:
Changes `SoStringA` type to `QString`

Fixes commit 911ef782f97b13d9454e4ef3f61dd020967802ce


itsmattkc:
Thanks for the PR, but I ended up having to rewrite this function anyway because it didn't work as-is.

PR #1153, opened 2020-05-08
Clean histogram - DO NOT MERGE

sobotka:
Review only. Issues:
 * Could be more performant. Not sure where.
 * Isn't scaled to a subset of the region for text.
 * The scaling of the Y axis is still and unknown. Normalizing with values in a shader is tricky. Could be feasible to add another texture and store the sum in there. Currently uses a power function.
 * Expose some UI things.


sobotka:
Maybe you can look into the scaling issues on this. In theory, it should work out that the sum of the rows works out for an image such that it can reveal when 50% of an image is a constant code value. I wasn’t able to bolt it down wholly, likely due to my shader knowledge. Maybe you can @itsmattkc?

PR #1154, opened 2020-05-08
Add Reset to Defaults button in Preferences

### LINKED ISSUE #1363: [UI] Add Reset to Defaults button in Preferences
<a href="https://github.com/ThomasWilshaw"></img></a> **Issue by [ThomasWilshaw](https://github.com/ThomasWilshaw)**
_Thursday May 07, 2020 at 20:51 GMT_
_Originally opened as https://github.com/sobotka/olive/issues/116_

----

**General TODO Summary**
Add a reset to defaults button in the preferences dialog.

**Additional Information / Notes**
Should this be per page of for the whole dialoge?

ThomasWilshaw:
Adds a reset button to preferences. At the moment I've just set up some of the signals and logic to get a rough plan.  Normal click resets the tab you're on, Ctrl click resets all.

In PreferencesDialog, when the reset button is pressed, we check which tab we are on and emit the corresponding signal to that object. The object will then reset all its options based of the config.

Does that sound reasonable/sensible? Are the casts I've done in the `connect()` correct or is there a better way of doing it?

Fixes #1363


ThomasWilshaw:
I've added a new QMap to the config that holds the original default values and an access method, it just makes a copy when SetDefaults() is called. We then reset values to this.

itsmattkc:
I think the cleanest way to do this (but would require a little more work to implement) would be for each Preferences tab, move all the lines that set the UI values from Config to a separate function that can take a "Config instance" as a parameter. Then in the constructor, call that function with `Config::Current()` to load the current config, then on resetting to defaults, run the same function again with a default Config instance.

ThomasWilshaw:
@itsmattkc The keyboard shortcuts page in preferences already has a reset button. Would you prefer I skipped that page or incorporated it into the new button?

itsmattkc:
I think it would make the most sense to incorporate into the new button, no real need for multiple reset buttons.

ThomasWilshaw:
Will do.

ThomasWilshaw:
Commits squashed and rebased to master. I think this is ready for review.

ThomasWilshaw:
I've updated this to match changes in `master`. I've removed some hardcoded indexes in `preferences.h` and added a function to find them automatically but I don't know if this is an improvement or not, there may be a better way. Also as the disc cache settings are now handled by `Diskmanager` they are no longer reset. Given it's not in the config I'm not sure what approach would be best here.

itsmattkc:
Looking over the code, I'm afraid there are a few issues:

* The default language in `Config` shouldn't be "en_US". Olive uses the system locale on startup (i.e. tries to match the language to the OS) unless a language is set specifically in the config. Setting it to "en_US" by default will break the auto-detection on startup.
* There shouldn't be a separate signal for every tab (e.g. `&PreferencesDialog::ResetGeneral` and `&PreferencesDialog::ResetAppearance`). It's going to make it a lot harder to add and remove tabs in the future if there's a lot of boilerplate code that needs to be added for each one. I notice you have separate `ResetDefaults()` functions for each tab, which is probably why you felt you had to do it that way, but it'd be better to have a virtual `ResetDefaults` function on the base `PreferencesTab` class so you don't have to manually identify what each tab is - you can just assume that every tab is a `PreferencesTab` and call `ResetDefaults` on it. With that, you can simply iterate over the `tabs_` array to reset defaults on all (see `PreferencesDialog::accept()` for an example) or just call reset on the `currentWidget()` (casting it to `PreferencesTab`) if the user is just resetting the current tab.
* Related to above, you shouldn't do string comparisons to translated user strings to identify each tab. String comparisons are very slow, but also since they're translated strings, they're unreliable. If a translation for whatever reason doesn't match all the `tr("General")` instances, the functionality will break. However if you do the above point, this one doesn't matter, since you won't need to identify tabs anyway.
* Also related to above, it's pretty wasteful to have a separate `QMessageBox` in every single tab. This also increases the maintenance cost dramatically because if we ever want to change that `QMessageBox` text, we'll have to change it in manually in every single tab. Instead, you should have that `QMessageBox` in `PreferencesDialog::RestoreDefaultSettings`, and use `QString::arg` with `list_widget_->currentItem()->text()` to automatically produce a message box that has the correct name in it. That would dramatically reduce the amount of code you need to change and the amount we need to maintain in the future.
* Minor point, but there shouldn't be any difference between `QString("")` and `QString()`. In fact, the former will be slower because it has to convert a `const char *` to a `QString` at runtime.

Finally, I hate to say this after you've spent quite a bit of time on it, but I think the question is worth asking: are there actually a lot of use cases for people resetting preferences to defaults in Olive? I can see maybe a handful of settings that people might want to reset (e.g. cache ahead/behind time or node colors maybe, which we can solve with the alt+click paradigm that already exists on the NodeParamView), but I'm not sure about the entire thing or even entire pages (i.e. how likely is it that a user would change the language or audio output device and then want to reset it to default)? This is an open question, people can feel free to chime in if they'd like. It just strikes me as perhaps not necessary for the majority of the settings that are there.

ThomasWilshaw:
Thanks for the review. I agree with all your points, I wrote the code ages ago and have learnt a lot since so if it was scrapped I wouldn't be at all bothered, equally I'll happily update it if that's wanted. In terms of actaully keeping I can see your point but I do think it,s worth being able ot reset most things (in terms of sound if you change hardware you might want ot reset to bug fix it not being picked upor something?). What would you think to making all UI items resetable including drop downs and check boxes?

itsmattkc:
I think we could potentially look into resetting individual elements at some point. For now I'll close this.

PR #1155, opened 2020-05-11
Waveform Updates

sobotka:
Should be safe for merging @itsmattkc.

Fixes:
 * Values exceeding the encoded range.
   Future may account for a certain range
   above and below current values.
 * Minor calculation problem that prevented
   the above fix from working properly.
 * Streamlined the shader code.


sobotka:
Added in the vertex shader, and made some minor modifications. Calling this done.

PR #1159, opened 2020-05-16
Histogram updates

sobotka:
Needs some review. Works. Scaled. Needs
some further UI love. Would like to move this
to a geometry shader to vastly improve performance.

### DIFF
--- a/app/shaders/rgbhistogram.frag
+++ b/app/shaders/rgbhistogram.frag
@@ -1,0 +1,37 @@
+#version 150
+
+uniform sampler2D ove_maintex;
+uniform vec2 ove_resolution;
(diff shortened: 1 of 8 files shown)


itsmattkc:
Looks like the PR is missing `shaders/rgbhistogram.vert`

sobotka:
Whoops. Put the commented geometry shader in instead of the proper vertex shader.

sobotka:
Ok. Updated and squashed. Calling this done for now until the geometry shader approach lands. Probably @ThomasWilshaw could look at adding a toggle to not draw the lines and text, as OpenGL is a slow pig for that.

PR #1160, opened 2020-05-19
Save state of certain floating panels

ThomasWilshaw:
Fixes sobotka/Olive#81

I just want to check I'm heading in the right direction with this. I've only implemented the scopes so far but I assume the curve panel should be more or less the same. I've added a couple of functions to Scopebase so I can access the type of scope and that will be saved as an attribute. It's all a bit messy at the moment, would you prefer some of the code was moved to a more appropriate place such as Scopebase?

### DIFF
--- a/app/panel/scope/scope.cpp
+++ b/app/panel/scope/scope.cpp
@@ -68,6 +68,12 @@ ScopePanel::ScopePanel(QWidget* parent) :
 void ScopePanel::SetType(ScopePanel::Type t)
 {
   scope_type_combobox_->setCurrentIndex(t);
+  type_ = t;
+}
+
+ScopePanel::Type ScopePanel::GetType()
+{
+  return type_;
 }
 
 QString ScopePanel::TypeToName(ScopePanel::Type t)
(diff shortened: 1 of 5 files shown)


itsmattkc:
I wouldn't recommend doing this one, it's actually deceptively complex. For example, the `sequence_viewer_panel_ ` is not the only panel that's allowed to spawn scopes. Any ViewerPanel can do it, including ViewerPanels that are spawned later by the user (which also need to be restored with each ScopePanel matching correctly to the viewer that spawned it). Same with any spawned CurvePanels that should match to the NodeInput they were spawned from.

ThomasWilshaw:
Okay no problem, it does work at the moment so hpefully it might be of some use down the line. Is it actually possible to spwan new viewers yet? Inotice the context menu for the footage view seeems to be disabled at the moment so you can't spawn a scope from there.

itsmattkc:
No, but it's on the list at sobotka/Olive#115.

The footage viewer should be able to spawn scopes, it's a bug if not

PR #1162, opened 2020-05-27
Fixes Issue 124 Ingestion Cache Multi Select Not Honoured

ThomasWilshaw:
Fixes sobotka/olive#124 (unless I've misunderstood something)

Multiple clips can now be proxied at once by selecting them all then setting the proxy level as usual. I've not been able to extensively test this as my laptop can only proxy #2 clips at once without dying, so that is probably worht doing before commiting. As an aside, proxying seems to use an awful lot of RAM.


itsmattkc:
Thanks for the PR, but I've ended up rewriting proxies into a proper pre-cache and addressed this issue in the process

PR #1163, opened 2020-05-28
Issue 72 Save footage proxy information to project

ThomasWilshaw:
Fixes sobotka/olive#72

Proxy  level is added to the `.ove` file in the stream section (similar to how `colorspace` is stored) and is retireved/set up. The only issue is how to tell the user if a proxy no loger exists? At the moment I'm using qInfo() but I suppose it would probaly be better to have a pop up. The issue is loading a project is done on a new thread (I think) which makes passing info back and forth a little tricky. Would it be reasonable to add a new function to `TaskDialog` , similar to `TaskFailed` the returns if a task completes but has non critical errors? Could be helpful with sobotka/olive#101. Might be over thinking it though.

Also the extending of `LoadCustomParameters` could probably be done in a nicer way, I feel the `break` statement could cause issues down the line. What are your thoughts?

### DIFF
--- a/app/project/item/footage/imagestream.cpp
+++ b/app/project/item/footage/imagestream.cpp
@@ -54,6 +54,7 @@ void ImageStream::LoadCustomParameters(QXmlStreamReader *reader)
   while (XMLReadNextStartElement(reader)) {
     if (reader->name() == QStringLiteral("colorspace")) {
       set_colorspace(reader->readElementText());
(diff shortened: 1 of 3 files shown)


itsmattkc:
Thanks for the PR, but with the proper pre-cache in 9095dfa463ff2dcecb7f801c8446a289f9918ed3 this has proved unnecessary.

PR #1164, opened 2020-06-01
Issue 70 Implement scrolling and panning in viewer

ThomasWilshaw:
Fixes sobotka/Olive#70

I've added basic translation functionality to the viewer. Assuing this is the right aproach I think there's some discussion to be had about specific functionality. For example, Ctrl clicking resets the image to the center of the view but doesn't affect the scaling, which is posibly not ideal?

Having just re read the original issue, do you only want this when the viewer is zoomed in?

Had to separate zoom and translate matrices to stop the zoom being applied to itelf. This also required adding separate get/set functions for each matrix and a GetCombinedMatrix() funtion.

Added those functions to other files where needed but may be wrong. Need to look at export.cpp. Gizmos are also not quite right yet either.

There's probaly a few improvements that could be made to the feel as well.


ThomasWilshaw:
Panning is now only possible if the viewer is zoomed in.

itsmattkc:
Indeed only panning when there's parts of the canvas offscreen, however my plan was to show scrollbars when zoomed in to make panning possible without the middle mouse (as opposed to Ctrl+drag). It should also drag with the primary mouse button when the hand tool is selected. Those behaviors would keep it in line with the other widgets.

ThomasWilshaw:
Okay no problem, is it correct to be doing all this through the OpenGL matrix?

PR #1167, opened 2020-06-23
Node dock

ThomasWilshaw:
Initial implementation of node bookmarking. Does this approach seem sensible? I've not massively tested it yet.

Fixes sobotka/olive#123


itsmattkc:
Thanks so much for the PR, I've used a number of the ideas from here and implemented it into the current master

PR #1168, opened 2020-07-01
Attempt to fix the current Travis Linux build issue

ThomasWilshaw:
I think this should fix the build issue but if the Travis build stiill fails I'll delete it.

### DIFF
--- a/app/widget/curvewidget/curveview.cpp
+++ b/app/widget/curvewidget/curveview.cpp
@@ -23,6 +23,7 @@
 #include <QScrollBar>
 #include <QMouseEvent>
 #include <QtMath>
+#include <limits>
 
 #include "common/qtutils.h"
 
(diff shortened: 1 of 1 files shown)


ThomasWilshaw:
Seems to build successfully on all the platforms now.

itsmattkc:
I've tended to use `#include <cfloat>` to address this in the past. Perhaps using STL is better though?

To make things consistent, I'd say you should either use `<cfloat>` in this PR or change the others to `<limits>` as well (there are only two other source files where this should be necessary).

ThomasWilshaw:
As far as I can tell there isn't much difference except that `<limits>` might be the more "proper" C++ way as it uses templates not macros (I could be completely wrong there though). That being said, `<limits>` just includes `<cfloat>` anyway (at least on Windows) so that seems a bit arbitary. I suppose `<cfloat>` is a bit more lightweight so I'll change it to that. I should really have checked what was already being done so sorry about that.

itsmattkc:
Depends on the compiler. On MSVC for example, you don't need to `#include <memory>` for smart pointers, but you do on GCC. I think some compilers just include more standard headers by default.

ThomasWilshaw:
Thanks, that's good to know.

PR #1169, opened 2020-07-03
Select node when right clicking

ThomasWilshaw:
Previously right clicking on a node would not select it. This often led to annoying problems of accidentally changing the label or copying the wrong node.

The new behaviour deselects all other nodes and selects the current node if you right click on it. However if the shift key is held whilst right clicking then any other currently selected nodes stay selected.

This is similar to the behaviour in most file browsers etc. so should be familiar to users.

Does this seem resonable?


itsmattkc:
Thanks for the suggestion, I ended up re-using your code in NodeView so that it can apply to both NodeViewItems and NodeViewEdges

PR #1171, opened 2020-07-26
Add S-Log2/S-Gamut3 and S-Log3/S-Gamut3.cine to OCIO config

pwnage101:
Gamut transformation matricies for S-Gamut3 and S-Gamut3.cine were pulled verbatim off of an .xlsx document published at Sony's website and archived at archive.org:

https://web.archive.org/web/20171116200834/http://community.sony.com/sony/attachments/sony/large-sensor-camera-F5-F55/12359/3/S-Gamut3_S-Gamut3Cine_Matrix.xlsx

The log->linear LUTs were generated by a python script that I wrote in a Jupyter Notebook (which I can share if you desire).  I made sure the input range of the LUT covered the "full" video (i.e. IRE -7.3% to 109.5%), and the output is always "scene linear reflection" rather than "scene linear IRE" (so I would multiply by 0.9 whenever necessary).  I've been able to find absolutely **no** information on the internet regarding the difference between scene linear "IRE" and "reflection" besides the fact that they are proportional and the ratio is apparently 0.9:1.  The only reason I have chosen reflection is because I actually reverse-engineered the V-Log LUT that's already committed to this repository and have determined it uses the "reflection", so I opted for the status-quo.

* S-Log2 specification: https://web.archive.org/web/20150718060914/https://pro.sony.com/bbsccms/assets/files/micro/dmpc/training/S-Log2_Technical_PaperV1_0.pdf
* S-Log3 specification: https://web.archive.org/web/20170223030529/http://community.sony.com/sony/attachments/sony/large-sensor-camera-F5-F55/12359/2/TechnicalSummary_for_S-Gamut3Cine_S-Gamut3_S-Log3_V1_00.pdf

Testing
---

I tested it and it looks fine.  No clue how to more rigorously test this, but I don't have any more time to work on this anyway.

By the way, I just donated $100---good luck!


sobotka:
Nice work. Given Olive is still in testing, would it help you to have the SLog / SGamut combination in the repository? If so, I can add them to the testing configuration.

Typically reflection is the target for scene linear workflows.

pwnage101:
At the moment I have no need for S-Log1, or Olive in general. Anyway, blender remains my primary NLE until Olive becomes semi-stable.

sobotka:
The twin Arri definitions are included for testing purposes, as Arri provides source footage that can be considered a reference standard for testing. If there are people out there with specific camera configurations that are engaged in testing, it would make sense to include those in the default installation.

Given this doesn’t seem to be the case, keeping the configuration down to the testing footage base set is sane.

PR #1172, opened 2020-08-11
Set up and implement UI controls for scopes

ThomasWilshaw:
I've added another `QStackedWidget` to the scope panel that can hold the UI controls for the selected scope. The controls themselves are created in the scope classes (`WaveformScope` and `HistogramScope`) and then a pointer is passed to the scope panel but I'm not sure if this is the best way to do it. I could put them into a new classes and hook it all up with connections but they don't really seem big enough to be worth separating out. What do you think?

I've also re implemented the same controls as I'd added in PR 1161 and PR1144


ThomasWilshaw:
Closing this for now as it's very out of date and not a priority

PR #1179, opened 2020-09-14
Prefer curl on OSX

rnkn:
I'm using the Travis CI scripts to build locally (the build instructions on the website seem to refer to 0.1.0). Maybe at some point the CI scripts could be used for both purposes, in which case it would be a tiny bit easier to prefer curl on OSX as wget is not available by default.

### DIFF
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -26,7 +26,7 @@ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
     macdeployqt $BUNDLE_NAME
 
     # Fix other deps that macdeployqt missed
-    wget -c -nv https://github.com/arl/macdeployqtfix/raw/master/macdeployqtfix.py
+    curl -sLOC - https://github.com/arl/macdeployqtfix/raw/master/macdeployqtfix.py
     python2 macdeployqtfix.py $BUNDLE_NAME/Contents/MacOS/Olive /usr/local/Cellar/qt5/5.*/
 
     # Fix deps on crash handler
--- a/.travis/script.sh
+++ b/.travis/script.sh
@@ -26,7 +26,7 @@ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
     macdeployqt $BUNDLE_NAME
 
     # Fix other deps that macdeployqt missed
-    curl -sLOC - https://github.com/arl/macdeployqtfix/raw/master/macdeployqtfix.py
+    curl -fLOSs --retry 3 https://github.com/arl/macdeployqtfix/raw/master/macdeployqtfix.py
     python2 macdeployqtfix.py $BUNDLE_NAME/Contents/MacOS/Olive /usr/local/Cellar/qt5/5.*/
 
     # Fix deps on crash handler


.travis/script.sh
```diff
@@ -26,7 +26,7 @@ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
     macdeployqt $BUNDLE_NAME
 
     # Fix other deps that macdeployqt missed
-    wget -c -nv https://github.com/arl/macdeployqtfix/raw/master/macdeployqtfix.py
+    curl -sLOC - https://github.com/arl/macdeployqtfix/raw/master/macdeployqtfix.py
```
Simran-B, line 29: > Use "-C -" to tell curl to automatically find out where/how to resume the transfer. It then uses the given output/input files to figure that out.

What purpose does this option serve here?

rnkn: It's just replicating the existing behaviour of `wget -c -nv`.

Simran-B: Hm, I wonder if wget retries a failed transfer. In that case it would make a lot of sense. But does curl can do that too?

itsmattkc:
We're actually in the process of migrating our CI to GitHub Actions. We can keep this in mind but the PR itself might be unnecessary. I'll see what other @Simran-B thinks though

Simran-B:
I say let's just merge it until the new CI is up and running. I'll substitute all wget usage with curl in the new scripts.

.travis/script.sh
Simran-B: > Note that you don't need to specify this option if you just want the current invocation of Wget to retry downloading a file should the connection be lost midway through. This is the default behavior.

So wget does retry, 20 times by default. Curl does not:

> If a transient error is returned when curl tries to perform a transfer, it will retry this number of times before giving up. Setting the number to 0 makes curl do no retries (which is the default)

There are also some remarks in the wget man pages about `-c` in v1.7+ which can lead to a download failure. Thus I think it should not be used and I'd drop the `-C -` for curl. We might want to add `--retry 3` or something though.

rnkn: It's more so curl won't download the file repeatedly when performing subsequent builds.

Simran-B: Actually, this does not replicate wget's behavior... In case of an HTTP error, curl will just save the error reply to a file and return 0 as exit code (success). wget fails in such a case. I think we should thus use the following replacement:

```suggestion
    curl -fLOSs --retry 3 https://github.com/arl/macdeployqtfix/raw/master/macdeployqtfix.py
```

- `-f` fail on error
- `-L` follow redirects
- `-O` output to file
- `-S` print errors
- `-s` but be silent otherwise

rnkn: You forgot the `-C` flag to avoid redownloading over and over btw.

Simran-B: I actually removed it on purpose, because it doesn't do anything here. On each CI run, a clean environment is used, hence all files which are not part of the build image need to be downloaded.

itsmattkc:
Is this ready for merging?

rnkn:
Sure. I added the -C flag because the script can double its usefulness for both CI and building locally. Removing it just makes things a tiny bit more difficult for no benefit. But main thing is not requiring wget.

Simran-B:
For local usage I have nothing against -C, it's just that this script won't work locally as-is unless you set the Travis-specific environment variables anyway. I updated the upcoming Linux CI script to use curl over wget BTW.

rnkn:
All good. I was gonna submit some followup PRs to make things work locally e.g. `[[ "$TRAVIS_OS_NAME" == "osx" || $(uname -s) == Darwin ]]` but I'll sit tight and see how you guys implement things :)

PR #1184, opened 2020-09-15
Fix/improve cmake crashpad discovery & compilation under Unix

Simran-B:
- `obj/` folder is in `out/Default` if we follow the crashpad compile instructions, thus we need to search for that
- Crashpad needs memfd_create(), which is available from glibc 2.27, but CentOS 7 comes with 2.17. Therefore we need to link the compat component, which in turn requires libdl.so to be linked (`-ldl`). CMake has a built-in `${CMAKE_DL_LIBS}` we can use. Not sure if this needs to be added conditionally (not on Windows), or if it's empty anyway.
- `client/client` etc. confused me quite a bit, because it looks like a path. Uses a cleaner implementation without regexes now.

TODOs:

- [x] Test on Linux/CentOS 7 with Clang:
  `cmake .. -G Ninja -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCRASHPAD_LOCATION=/opt/olive/crashpad/crashpad`
- [x] Test on Linux/CentOS 7 with GCC
- [ ] Test on macOS
- [ ] Test on Windows
- [ ] ~GCC toolchain option:~
  Crashpad is compiled with Clang by default. Linking Olive with the GCC toolchain against it fails with many symbol errors, so we need to either find a compatibility flag in GCC (ld) or figure out how to make gn compile Crashpad with GCC.
  **//update:** The issues were causes by missing libs, GCC Olive build links fine again clang crashpad build.

Questions:

- Which of the Crashpad build artifacts do we need to ship with Olive? Just the crashpad_handler binary? Is everything else compiled into Olive? (libclient.a etc. under Unix)
- The handler should be in the same folder as the olive-editor binary I suppose? And not be hardcoded to Windows (`.exe`):
  `base::FilePath handler(TO_BASE_STRING_TYPE(QDir(exe_dir).filePath(QStringLiteral("crashpad_handler.exe"))));`
  https://github.com/olive-editor/olive/blob/master/app/common/crashpadinterface.cpp#L50


app/CMakeLists.txt
```diff
@@ -182,6 +182,9 @@ if (GoogleCrashpad_FOUND)
     ${OLIVE_TARGET}
     PRIVATE
     ${CRASHPAD_LIBRARIES}
+    # TODO: UNIX only? Or is it empty on Windows anyway?
+    # HACK: crashpad compat lib needs libdl.so (-ldl)
```
Simran-B, line 186:
```suggestion
    # Crashpad compat lib needs libdl.so (-ldl)
```

itsmattkc:
I actually added the `out/Default` search in my local copy but forgot to push it. The rest of the changes appear to be cleanup and Linux support I'm guessing?

To answer your questions, Crashpad should be statically built into Olive, so yes the only thing that needs to be packaged is the `crashpad_handler`.

Yes the code currently only references the Windows executable but we'll eventually extend that.

Simran-B:
The issues with clang crashpad + gcc olive were apparently be caused by something I messed up in a  cmake script, which led to cmake not finding the crashpad libs. So it shouldn't be necessary to compile crashpad with gcc for gcc olive builds. A different toolchain could be chosen via env vars however (CC, CXX, AR)

Simran-B:
The crashpad location is discovered automatically, but I'm not actually sure why. Just because it's in `/usr/local` now? Curiously, this wasn't the case for breakpad until I added breakpad to PATH_SUFFIXES...

`cmake --install .`  copies crashpad_handler and minidump_stackwalk to the bin folder under Linux. Under macOS these files should be copied into the bundle (untested). On Windows we don't make use of the install target yet.

PR #1186, opened 2020-09-16
Windows CI improvements

Simran-B:


.github/workflows/ci.yml
```diff
@@ -23,118 +23,170 @@ jobs:
     steps:
     - name: Set package name
       shell: bash
-      run: echo "::set-env name=PKGNAME::$(echo Olive-${GITHUB_SHA::8}-Windows-x86_64)"
-      
+      run: |
+        echo "::set-env name=PKGNAME::$(echo Olive-${GITHUB_SHA::8}-Windows-x86_64)"
```
Simran-B, line 27: In a batch script, this following should work: `%GITHUB_SHA:~0,8%`

.github/workflows/ci.yml
```diff
@@ -1,142 +1,195 @@
...
 
     - name: Build
-      working-directory: ${{runner.workspace}}/build
+      working-directory: ${{ runner.workspace }}/build
       shell: bash
-      # Execute the build.  You can specify a specific target with "--target <NAME>"
-      run: cmake --build . --config $BUILD_TYPE
+      # Execute the build. You can specify a specific target with "--target <NAME>"
+      run: |
+        cmake --build . --config ${{ matrix.build-type }}
```
Simran-B, line 113: I doubt that `--config` is necessary here:

> --config <cfg> = For multi-configuration tools, choose <cfg>.

MSVC is multi-config, in which case `-DCMAKE_BUILD_TYPE` doesn't need to be set in the configure phase because it is decided in the build phase with `--config`. But Ninja is single config AFAIK, so only `-DCMAKE_BUILD_TYPE` is needed.

.github/workflows/ci.yml
```diff
@@ -1,142 +1,195 @@
...
-      working-directory: ${{runner.workspace}}/build
+      working-directory: ${{ runner.workspace }}/build
       shell: cmd
-      run: mkdir olive-editor && copy app\olive-editor.exe olive-editor && copy app\olive-editor.pdb olive-editor && windeployqt olive-editor\olive-editor.exe && copy %GITHUB_WORKSPACE%\bin\*.dll olive-editor && copy %GITHUB_WORKSPACE%\out\Default\crashpad_handler.exe olive-editor
-    
+      run: |
+        mkdir olive-editor
+        copy app\olive-editor.exe olive-editor
+        copy app\olive-editor.pdb olive-editor
+        windeployqt olive-editor\olive-editor.exe
```
Simran-B, line 130: Technically, we should also run deploy for olive-crashhandler.exe (though it shouldn't be necessary because olive-editor.exe uses a superset of libraries).

.github/workflows/ci.yml
```diff
@@ -1,142 +1,195 @@
...
-      working-directory: ${{runner.workspace}}/build
+      working-directory: ${{ runner.workspace }}/build
       shell: cmd
       env:
-        GH_AUTH_KEY: ${{secrets.GH_AUTH_KEY}}
-      run: curl -fLSs --retry 3 https://github.com/google/breakpad/blob/master/src/tools/windows/binaries/dump_syms.exe?raw=true > dump_syms.exe && dump_syms app\olive-editor.pdb > olive-editor.sym && curl -X POST -F symfile=@olive-editor.sym -F key="%GH_AUTH_KEY%" https://olivevideoeditor.org/crashpad/symbols.php
-    
+        GH_AUTH_KEY: ${{ secrets.GH_AUTH_KEY }}
+      run: |
+        curl -fLSs --retry 3 https://github.com/google/breakpad/blob/master/src/tools/windows/binaries/dump_syms.exe?raw=true > dump_syms.exe
```
Simran-B, line 155: Curl's `-O` option isn't very smart lol (dump_syms.exe_raw=true)

.github/workflows/ci.yml
itsmattkc: I didn't think so either, but that's just what the GitHub Actions template had and I didn't bother to change it

.github/workflows/ci.yml
itsmattkc: Yeah I'm fairly sure `olive-crashhandler.exe` isn't used anymore (though jury's out on whether it'll be used again in the future)

.github/workflows/ci.yml
Simran-B: In PowerShell:

```
echo "::set-env name=PKGNAME::$($env:GITHUB_SHA.substring(0, 8))"
```

.github/workflows/ci.yml
Simran-B: CMake actually supports `-G "Ninja Multi-Config"` to generate multiple .ninja files

https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html

Calling ninja and cmake correctly looks rather convoluted however.

PR #1207, opened 2020-09-20
Issue template: <details> for backtrace

Simran-B:
Make the Crash template not show the backtrace by default, as suggested by ZoomTen.

Use HTML comment for explanation of what the commit hash is, so that it doesn't end up in the ticket und is still visible to the user when reporting.


itsmattkc (commented):
Seems good, might even want to add the extra spacing and commit hash changes to the rest of the templates

itsmattkc (approved):

Issue #1190, opened 2020-09-17
[EDIT] Minor visual glitches in timeline


klwemu:
**Commit Hash**

**Platform**
Win10

**Summary**
When you delete a clip wich is a bit longer displayed on screen (does not depend on the real clip time length) a part is visually not deleted.

**Additiona
![Delete Clip](https://user-images.githubusercontent.com/3965097/93452061-4277e700-f8d8-11ea-8e90-ce3d7e10f8d1.png)
l Information / Output**

ThomasWilshaw:
Please provide a commit hash. Thanks.

klwemu:
Commit hash:
007b9f2

itsmattkc:
I've seen this before. I believe it's bugs in Qt's update system that we can workaround by setting the TimelineView to `QGraphicsView::FullViewportUpdate`. That's technically less optimized, but should address such glitches.

PR #1214, opened 2020-09-22
[UI] Fix keyframe icons not redrawing correctly

ThomasWilshaw:
When changing between nodes the keyframes in the keyframe editor did
not always update correctly. This is the same fix as in 9c488d9 / https://github.com/olive-editor/olive/issues/1190#issuecomment-694252893


itsmattkc:
Hmm if the QGraphicsViews are this unreliable, perhaps we should bite the bullet and set this for all of them (both TimelineView and KeyframeView inherit from TimelineViewBase). Perhaps you could remove the `setViewportUpdateMode` call from both KeyframeViewBase and TimelineView, and place it into TimelineViewBase to address all of them instead? (that will fix views like GraphView and any future derivatives too)

itsmattkc:
Done in a313fb525c5df38534ed69cd9f7c8bf6ab9803c5

Issue #1218, opened 2020-09-23
[EXPORT] export dialogue doesn't always show up


jarro2783:
**Commit Hash** 

d196689c2d4288

**Platform**

Linux

**Summary**

The export dialogue doesn't show up under some circumstances when I expect it to.

1. Open a project.
2. Export without clicking anything else in the UI.

The dialogue will not show up here, and no message is shown.

If you press play in the viewer, then subsequent export works.

**Additional Information / Output**

Simran-B:
Confirmed on Windows 10 d5595ccd

If you start Olive and immediately click _File > Export_, then nothing happens. Same if you open a project file and try to export. If you focus the timeline panel or click on a clip, then the export dialog shows up as expected.

ThomasWilshaw:
I think we just need a bit more logic and a error dialog in `Core::DialogExportShow()`. It currently just returns if there is no recently used `TimeBasesPanel`.

Simran-B:
While it's an improvement to show an error instead of silently failing, it does not seem right to me that there is no recently used time based panel / active sequence in the project load case. Can't this be set on load?

ThomasWilshaw:
What would you do in the case of loading a project with multiple sequences?

jarro2783:
Currently when I open a project, it already shows the first frame of the video in the preview window. So it seems reasonable for that video to be "selected", and be the one exported.

For multiple sequences it's probably reasonable enough to pick the one that appears first in the list. As long as what you see in the preview is the one exported, that seems intuitive enough to me.

ThomasWilshaw:
I've updated this so it falls back to the timeline if there is no recently selected time based panel.

itsmattkc:
Should be fixed in fd143f1d519cf41f7770d0d3849e0daaadc5bc91

PR #1220, opened 2020-09-23
Fix Issue #1218

ThomasWilshaw:
Add an error message if there is no active sequence.

Selected a Sequence panel as opposed to a generic time based panel as
we will only ever export a sequence. Also makes error logic slightly
easier.

### DIFF
--- a/app/core.cpp
+++ b/app/core.cpp
@@ -355,18 +355,23 @@ void Core::DialogProjectPropertiesShow()
 
 void Core::DialogExportShow()
 {
-  TimeBasedPanel* latest_time_based = PanelManager::instance()->MostRecentlyFocused<TimeBasedPanel>();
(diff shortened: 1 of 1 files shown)


itsmattkc:
I would prefer it to still be a TimeBasedPanel in some capacity so that people can select a timeline panel and export from that too, but there should definitely be an error message as you've written in for this scenario.

ThomasWilshaw:
Okay no problem, I'm going to change it a bit as well based of feedback in the initial issue.

ThomasWilshaw:
I was a little confused with the difference between a sequenceVewer and a timeline. Am I correct in thinking the sequence viewer is the viewer window panel and the timeline panel is the actual timeline? (seems obvious now I say it...)

itsmattkc:
Indeed, SequenceViewer is a derivation of Viewer that displays the active sequence and Timeline is the timeline.

ThomasWilshaw:
@itsmattkc How does this look?

itsmattkc:
Yep look good

PR #1244, opened 2020-09-29
Fix issue#1241

### LINKED ISSUE #1241: [CRASH] Removed Audio Node
**Commit Hash** 
610f4c54

**Platform**
Windows

**Summary**
Olive crashed when I imported a clip into a sequence, removed the audio node, and tried to move the clip.

**Steps to Reproduce**

1. Add Video with audio track to a sequence.
2. Delete the audio node from the node editor
3. Try to move the remaining video clip.

<details><summary><strong>Backtrace</strong></summary><pre><code>

</code></pre></details>

[crash.txt](https://github.com/olive-editor/olive/files/5293730/crash.txt)

**Additional Information**

Also, I don't think the crash reporter works. It kept saying "Failed to send error report."

ThomasWilshaw:
Fixes #1241 

When a Block node was deleted its linked blocks were not unlinked
which caused a crash when the user tried to select a linked block.
This fixes the issue by ensuring all links are removed when a Block
node is delete.

Undo/redo is also supported.

This does fix the problem but I'm not 100% sure this code is in the correct place.

### DIFF
--- a/app/widget/nodeview/nodeviewundo.cpp
+++ b/app/widget/nodeview/nodeviewundo.cpp
@@ -154,6 +154,16 @@ void NodeRemoveCommand::redo_internal()
 
   // Take nodes from graph (TakeNode() will automatically disconnect edges)
   foreach (Node* n, nodes_) {
+    // If the node is a block, unlink any linked blocks before removing
+    if (n->IsBlock()) {
+      Block *b = static_cast<Block *>(n);
+      if (b->HasLinks()) {
+        linked_blocks_.insert(b, b->linked_clips().toList());
+        foreach(Block * link, b->linked_clips()) {
+          b->Unlink(b, link);
+        }
+      }
+    }
     graph_->TakeNode(n, &memory_manager_);
   }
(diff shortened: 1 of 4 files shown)


itsmattkc:
Here's the thing, the code _does_ unlink when you delete a block, but only the timeline does it because I hadn't considered people deleting blocks in the node graph: https://github.com/olive-editor/olive/blob/master/app/widget/timelinewidget/timelinewidget.cpp#L497

Unlinking in NodeRemoveCommand instead probably is the most fool-proof thing to do, in that that'll affect all node removals including the timeline, but it might be best if you can reuse the BlockUnlinkAllCommand in NodeRemoveCommand.

ThomasWilshaw:
Okay will do, I guess `BlockUnlinkAllCommand` should stay in the timeline undo file though?

ThomasWilshaw:
I've replaced my code with the `BlockUnlinkAllCommand` and removed all the other block unlink commands so nothing is happening twice.

PR #1262, opened 2020-10-09
Fix #1252

### LINKED ISSUE #1252: [NODES] Viewer Node + Track Node Crash
ebd9e271
Windows 10 64 Bit

Deleting the line that connects from track to samples in the viewer node causes a segfault

This can be easily reproduced if you create a new sequence, and delete/disconnect said line connecting track to samples.

image

ThomasWilshaw:
Checks if the SampleBuffer is valid before trying to write it to a
PCM cache file. If it's not, write a blank PCM of the correct lenght
and print a warning.

Fix #1252

### DIFF
--- a/app/render/audioplaybackcache.cpp
+++ b/app/render/audioplaybackcache.cpp
@@ -75,8 +75,11 @@ void AudioPlaybackCache::WritePCM(const TimeRange &range, SampleBufferPtr sample
     segment_length_ += seg_sz;
   }
 
-  // Convert to packed data, which is what we store on disk
-  QByteArray a = samples->toPackedData();
+  QByteArray a;
+  if (samples) {
+    // Convert to packed data, which is what we store on disk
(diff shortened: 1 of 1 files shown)


itsmattkc:
This PR might be out of date since I just rewrote the audio cache, however the fix is probably still appropriate if you'd like to rebase

ThomasWilshaw:
I've rebased and updated this and it seems to work fine.

itsmattkc:
Sorry, I read back through the function and realized it already fills in remaining bytes with zeroes so the fill isn't necessary. Also I think it's probably fine for that function to "officially" accept null sample buffers as a means to just write silence, so really it's just the if statement that's necessary - no need for the warning.

ThomasWilshaw:
I see what you mean, I think I misunderstood what the function was doing. I'll do that now.

itsmattkc:
Okay awesome, thanks

PR #1263, opened 2020-10-10
Fix #1253 (Timeline does not update properly on new sequence)

### LINKED ISSUE #1253: [UI] Timeline does not update properly on new sequence
ebd9e271

Windows 10 64 Bit

You can add extra tracks with ease by just adding clips in other areas of the timeline, but once you have multiple tracks and decide maybe you wanna go delete that sequence and create a new one, the timeline UI does not update properly. 

Steps to Reproduce:
1. create a new sequence
2. add as many tracks as you want, as long as there is more than 2
3. Delete your sequence
4. Make a new sequence

Result: you get 2 tracks, but the track lables and controls do not update properly back to 2 tracks only. 
image

ThomasWilshaw:
@itsmattkc I don't think this is the fix you'll want but I can't quite figure out a better way of doing it. As far as I can tell the problem here is that the track lists indexes don't get properly updated when a track is deleted so the code in `TrackViewSplitter::Remove` messes up. If you uncomment the `foreach` loop in `TrackView::ConnectTrackList` and look at the print out you get you can see how bad it gets:

```
count: 4, index: 0 => 3

count: 3, index: 1 => 1

count: 2, index: 2 => -1

count: 2, index: 0 => 0

```

The function I have written gets around the problem but it seems quite hacky and it's probably better if the lists get updated. I do think the timeline needs to be cleared though when closing a sequence to fix #1253

### DIFF
--- a/app/widget/timelinewidget/timelinewidget.cpp
+++ b/app/widget/timelinewidget/timelinewidget.cpp
@@ -185,11 +185,6 @@ void TimelineWidget::Clear()
   // Emit that we've deselected any selected blocks
   SignalDeselectedAllBlocks();
 
-  foreach (TimelineAndTrackView* tview, views_) {
-    tview->track_view()->DisconnectTrackList();
-    tview->view()->ConnectTrackList(nullptr);
-  }
-
   // Set null timebase
   SetTimebase(0);
 }
--- a/app/widget/timelinewidget/timelinewidget.cpp
+++ b/app/widget/timelinewidget/timelinewidget.cpp
@@ -185,6 +185,11 @@ void TimelineWidget::Clear()
   // Emit that we've deselected any selected blocks
   SignalDeselectedAllBlocks();
 
+  foreach (TimelineAndTrackView* tview, views_) {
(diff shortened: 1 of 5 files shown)


ThomasWilshaw:
I think this is now working but I don't really understand this area of code enough to know if what I've done here is sensible or not.

itsmattkc:
I haven't had a chance to fully investigate this, but there actually is code intended to update the track indexes: https://github.com/olive-editor/olive/blob/master/app/node/output/track/tracklist.cpp#L227

Maybe you can figure out why that's not functioning as expected?

ThomasWilshaw:
The function you've linked to is only called when nodes are disconnected, however closing or deleting sequence doesn't actually appear to touch the nodes at all, it just deletes the panel.

itsmattkc:
Ahh you were right all along actually, the index given to the splitter corresponds to the index of the widget, not the index of the track. While they're syncronized in normal usage, when removing them in a for loop, the indexes will all shift whenever a widget is removed. I implemented essentially the same fix in `nodearchchanges`, but yes this is actually the correct solution.

PR #1264, opened 2020-10-11
Reset Timecode in Footage Viewer

ThomasWilshaw:
Resets the current timecode to 0 when the footage in the viewer is
closed.

Hides the end timecode of a viewer whenever the end timecode is equal
to 0.

### DIFF
--- a/app/widget/playbackcontrols/playbackcontrols.cpp
+++ b/app/widget/playbackcontrols/playbackcontrols.cpp
@@ -186,6 +186,10 @@ void PlaybackControls::SetEndTime(const int64_t &r)
   end_tc_lbl_->setText(Timecode::timestamp_to_timecode(end_time_,
                                                        time_base_,
                                                        Core::instance()->GetTimecodeDisplay()));
+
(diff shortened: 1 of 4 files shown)


itsmattkc:
I used some of this, but I didn't think the end timecode label should be hidden if the length is 0 (that would mean it would hide if the timeline was empty too which feels undesirable). I've otherwise implemented this in f137a1bbf348737d53058e73711e6b2951e54895, but thanks for pointing me in the right direction

ThomasWilshaw:
The reason I hid the timecode is that when Olive first opens the end timecodes are not visible. I think with the footage viewer it may be nice as you can be viewing things with different timebases (25, 29.97 etc.) so the format can change between a `:` and a `;`.

itsmattkc:
Yes I figured that was the idea, I just didn't think hiding it on 0 duration was the right approach. But I did make that behavior more consistent by always hiding both timecodes if no media is loaded

ThomasWilshaw:
Cool, makes sense. I hadn't had chance to check the actual commits yet.

PR #1267, opened 2020-10-13
Handle Missing Files

### LINKED ISSUE #1260: [CRASH] Changing location of imported files crashes Olive on project opening
**Commit Hash** 
0.2.0 ac4be090

**Platform**
Windows 10

**Summary**
If the location of the imported files is changed in a Project, Olive crashes opeining the project.

**Steps to Reproduce**

1. Launch Olive and go to: `File > Import...` and import a file (same with all formats)
2. Save the Project
3. Change the location of the imported file in step 1

Olive crashes consistently.

<details><summary><strong>Backtrace</strong></summary><pre><code>

</code></pre></details>

**Additional Information**

ThomasWilshaw:
Fix #1260 and fix sobotka/Olive#101

This is a basic fix for handling missing files on load. If a footage probe returns false the clip/nodes are created but with null pointers for there streams and the clip is marked as `kInvalid` in the project explorer. It would be good to also add a warning message and maybe some kind of visual indicator in the timeline and/or the viewer.

Also need to add a way to relink the footage.

### DIFF
--- a/app/codec/decoder.cpp
+++ b/app/codec/decoder.cpp
@@ -101,12 +101,16 @@ bool Decoder::ProbeMedia(Footage *f, const QAtomicInt* cancelled)
   // Check for a valid filename
   if (f->filename().isEmpty()) {
     qWarning() << "Tried to probe media with an empty filename";
+    f->set_status(Footage::kInvalid);
+    f->set_decoder(QString());
     return false;
   }
 
   // Check file exists
   if (!QFileInfo::exists(f->filename())) {
     qWarning() << "Tried to probe file that doesn't exist:" << f->filename();
+    f->set_status(Footage::kInvalid);
+    f->set_decoder(QString());
     return false;
   }
 
--- a/app/dialog/footageproperties/footageproperties.cpp
+++ b/app/dialog/footageproperties/footageproperties.cpp
@@ -30,8 +30,10 @@
 #include <QListWidget>
 #include <QCheckBox>
 #include <QSpinBox>
+#include <QFileDialog>
 
 #include "core.h"
+#include "codec/decoder.h"
 #include "render/colormanager.h"
 #include "streamproperties/audiostreamproperties.h"
 #include "streamproperties/videostreamproperties.h"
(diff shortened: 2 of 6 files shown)


ThomasWilshaw:
@itsmattkc this is more or less working in a fairly basic way (assuming you're happy with it), I'm just stuck on how to get the media nodes to link to re-linked footage. The media nodes need to somehow remember what their pointer and index _should be_. I was wondering about storing the stream pointer and stream index like this:

```
int stream_index;
quintptr stream_ptr;
```

in the media node so when a relink happens Olive can check through all media nodes and copy these values over. What do you think?

itsmattkc:
Thanks for taking a look at this, but I had some different ideas about its implementation.

Firstly I revamped the probe functions a fair bit in the `otio` branch, which I had been holding off on merging until the OTIO implementation was more complete, but might just be worth merging in now. The work there will unfortunately affect this PR quite a bit.

Secondly, I think the footage streams should be loaded regardless of whether the footage is found. That way we can keep any user settings pertaining to that footage even while its offline, and restore it whenever the footage does relink. That should address your second concern about whether the stream indexes should be stored in the media nodes.

I'm also not confident about there being a "path" entry in the footage properties dialog, but I might warm to it. I would implement some kind of QListWidget dialog after project load to show the user which files are missing and allow them to relink as Premiere and Pro Tools do.

ThomasWilshaw:
No problem. If you don't want to merge the OTIO branch yet I can redo this PR based off the OTIO branch. It'll probably be easier to just start it again either way (and should be quicker now I know a bit what I'm doing ;) ).

The problem I'm having with loading the streams regardless is that if the probe fails no streams are created so there is nothing meaningful that can be passed to the `XMLNodeData` in `Footage::Load`. I had to modify `Footage::stream()` to not throw a wobbly if `streams_` is empty so it now returns a `nullptr` in that case. I can't figure out what I need to pass to where to make sure the `MediaInput` node knows what stream it should eventually get. Does that make sense? That's why I wondered if having the media node store the footage and the stream index.

I agree a list widget appearing on load would be good. I think it would be good to show somewhere in the UI exactly what file a piece of Footage is referencing. I also think there should be a way to re-link any bit of footage whenever you want to (I think most editors can do this), would a new action in the project explorer right click menu be a better place for this?

ThomasWilshaw:
Okay I'll look into that. I notice all the footage Load code is commented out atm in `otio` is that being changed soon?

Cool, I'll move it to there then.

itsmattkc:
Ahh true, I need to fix that. I'll do that today, might even be able to get OTIO merged in too.

Simran-B:
@ThomasWilshaw Just in case that you wanna change this PR's base branch: Click on Edit at the top and select the `otio` branch, i.e. you don't have to create a new PR.

image

ThomasWilshaw:
Thanks I wasn't aware of that :) . It may be that refactoring my code is more work than just starting from scratch though.

Simran-B:
You may also overwrite your local branch to start from scratch, force-push it and - if necessary - change the base branch on GitHub.

ThomasWilshaw:
Ah okay, so as to keep the same GitHub PR going?

Simran-B:
Exactly. I believe the only situation where a new PR is unavoidable is if you try to rename the branch, which automatically closes the existing PR and creates a new branch IIRC.

ThomasWilshaw:
Closing as Matt has covered this in the OTIO branch.

PR #1268, opened 2020-10-13
Very basic color grading

JosefR:
This is a first start to add a color grading node for setting a color offset and the classic lift-gamma-gain setting. I have not yet spent any effort on UI work, this is basically just a simple fragment shader with minimal UI code. 

Please let me know if this is going in the right direction.

### DIFF
--- a/app/node/factory.cpp
+++ b/app/node/factory.cpp
@@ -31,7 +31,7 @@
 #include "generator/solid/solid.h"
 #include "generator/text/text.h"
 #include "filter/blur/blur.h"
-#include "filter/color/color.h"
+#include "filter/slopeoffsetpower/slopeoffsetpower.h"
 #include "filter/stroke/stroke.h"
 #include "input/media/video/video.h"
 #include "input/media/audio/audio.h"
@@ -205,7 +205,7 @@ Node *NodeFactory::CreateFromFactoryIndex(const NodeFactory::InternalID &id)
   case kBlurFilter:
     return new BlurFilterNode();
   case kColorFilter:
-    return new ColorFilterNode();
+    return new SopFilterNode();
   case kSolidGenerator:
     return new SolidGenerator();
   case kMerge:
--- a/app/node/factory.cpp
+++ b/app/node/factory.cpp
@@ -31,6 +31,7 @@
 #include "generator/solid/solid.h"
 #include "generator/text/text.h"
 #include "filter/blur/blur.h"
+#include "filter/color/color.h"
 #include "filter/stroke/stroke.h"
 #include "input/media/video/video.h"
 #include "input/media/audio/audio.h"
@@ -203,6 +204,8 @@ Node *NodeFactory::CreateFromFactoryIndex(const NodeFactory::InternalID &id)
     return new TimeInput();
   case kBlurFilter:
     return new BlurFilterNode();
+  case kColorFilter:
+    return new ColorFilterNode();
   case kSolidGenerator:
     return new SolidGenerator();
   case kMerge:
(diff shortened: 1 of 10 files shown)


sobotka:
Lift Gamma Gain doesn’t work. It’s scene linear.

Would need adjusting to Slope Offset Power, with a pivot.

ThomasWilshaw:
Slope:
`out = in * slope`

Offset:
`out = in + offset`

Power + pivot:
`power(in / pivot, power) * pivot`

Have I got that right @sobotka?

JosefR:
> Lift Gamma Gain doesn’t work. The working space in Olive is scene linear.
> 
> Would need adjusting to Slope Offset Power, with a pivot.

OK, my idea basically was to do the same as Nuke for example ist doing with it's grading node. I thought Nuke is also working in a scene linear space and they still have these settings. In the end it is just a mathematical operation, but of course the result will look different if it is applied to scene linear or a log space or the display space. 

If it still makes sense, I could try to replicate what Nuke or Natron are doing in their grading node:
https://natron.readthedocs.io/en/rb-2.3/plugins/net.sf.openfx.GradePlugin.html

sobotka:
No. The formulas are broken for radiometric-like RGB with LGG. The only one that works properly is SOP.

JosefR:
I just added slope-offset-power functions. 
[https://youtu.be/djmooVER95I](https://youtu.be/djmooVER95I)

ThomasWilshaw:
You need to add the pivot input (check the formulas above) and check you alpha handling, it should pass through, not be set to `1.0`

JosefR:
You are right, alpha values are passed through now. I am just not sure about the pivot. In the CDL descriptions I could find, it was not mentioned e.g. here: [https://en.wikipedia.org/wiki/ASC_CDL](https://en.wikipedia.org/wiki/ASC_CDL)

ThomasWilshaw:
Power functions effectively pivot around 1.0 (`power(1, n) =1`). By using the above formula you can set what code value the power function pivots around. From a user point of view this would be a single channel input value (i.e. all channels use the same value.

JosefR:
OK, I will prepare the pivot option. 

Another question would be, if there should be anything color grading related in one node or if there should be one node for each purpose. Then it would make sense to rename this node to "Slope Offset Power" and add another one for saturation or RGB curves. Having all in one node would have the advantage, that it could be optimized quite easy. I think you could compile the colour grading operations all into a 1D LUT which could speed up preview renders.

ThomasWilshaw:
I would definitely keep each function as a separate node. The whole point of a node is small separate building blocks. I think luts should be avoided where possible as they inevitably lose precision.

sobotka:
Technically slope is multiply, power is power, and offset is add. Would likely be good to start reusing code paths via the math node perhaps.

@itsmattkc is this sane at this point to bundle and work a node chain?

Saturation is much trickier as the primaries of the reference space need to be taken into account. This would lean on OCIO API calls, so it would likely make sense to put an Olive stub in to grab them. OCIO V2 might tidy this up a bit, therefore a stub is better than littering the nodes with actual API linkage.

Simran-B:
@sobotka Are you referring to grading ops? https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1102

sobotka:
Indeed.

Although I haven’t looked too closely to see if the “saturation” honours the working space primaries.

Worth a bit of a look.

JosefR:
@ThomasWilshaw Nodes could implement different options for preview or final rendering. For preview renders it might be acceptable to sacrifice quality for speed, if it's user-selectable. Especially for color grading, a responsive UI might be important.

PR #1269, opened 2020-10-14
Fix building without OTIO support

ThomasWilshaw:
I saw there was already some support for building `otio` _without_ OTIO support so got it working on Windows at least (I've not yet attempted building OTIO myself...). I don't really know CMake very well so I might not have done it properly.

### DIFF
--- a/app/codec/CMakeLists.txt
+++ b/app/codec/CMakeLists.txt
@@ -16,7 +16,9 @@
 
 add_subdirectory(ffmpeg)
 add_subdirectory(oiio)
-add_subdirectory(otio)
+if(OpenTimelineIO_FOUND)
+  add_subdirectory(otio)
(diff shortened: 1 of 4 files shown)


Simran-B:
I don't see a need to add a new variable `OTIO_FOUND`. The `find_package(OpenTimelineIO)` call automatically sets `OpenTimelineIO_FOUND`. You can just use that can't you?

`if (OpenTimelineIO_FOUND)`

ThomasWilshaw:
Ah okay, I got a bit confused with that. I'll update, thanks.

Simran-B:
CMake does have a lot of magic behavior that surprises you. But what can you do.

ThomasWilshaw:
Updated to use existing variable.

PR #1291, opened 2020-10-29
Remove old Travis and AppVeyor CI, update .gitignore

Simran-B:
Re .gitignore, I suppose it's fine to remove `effects/frei0r` as we don't use it anymore?
What was the `history` entry for? VS Code uses `.history`, which I added.


itsmattkc:
I can't remember what the `history` ignore is for, so I assume it's not important (might have been automatically added by something?)

Simran-B:
Haven't seen this one myself, so let's remove it for now. Added `/out/` to ignore VS CMake build dir, and restrict docs/ and build*/ to root folder.

Issue #1305, opened 2020-11-05
[CACHE] Clear cache menu entry


pgilfernandez:
**Commit Hash** 
de48eba7

**Platform** 
macOS 10.14.6

**Summary**
It would be nice to have a menu entry inside "Sequence" to clear the entire cache. Maybe it's something not useful while in production and when the software is robust, but in the meanwhile it would be nice to have a fast way to launch this task.

Thanks

**Additional Information / Output**

Simran-B:
So basically a quicker way to do this?

![image](https://user-images.githubusercontent.com/7819991/98237593-6b763900-1f65-11eb-8d45-ba706528617d.png)

pgilfernandez:
Yes, this is just what I'm looking for, without the need to close Olive. Have in mind that sometimes Olive crashes and it doesn't clear the cache...

Simran-B:
I guess it would make sense to change the feature to "Automatically clear disk cache on close and start", so that you can be sure that there are no leftovers, even if a crash occurred.

pgilfernandez:
Yes, that would make sense too, "on close and start".
But I still think the option to easily clear the cache on demand from the menu would be great (I'm experimenting cache issues and it would be handy... I guess more people will find it also handy)

Simran-B:
Here's a PR: #1306

I'm slightly concerned that the path it tries to delete is `''`. This is also the case with the existing dialog...

![image](https://user-images.githubusercontent.com/7819991/98245708-a0888880-1f71-11eb-8eb5-849362a6a112.png)

elsandosgrande:
> Here's a PR: #1306
> 
> I'm slightly concerned that the path it tries to delete is `''`. This is also the case with the existing dialog...
> 
> ![image](https://user-images.githubusercontent.com/7819991/98245708-a0888880-1f71-11eb-8eb5-849362a6a112.png)

I'm still seeing this to this day:
![image](https://user-images.githubusercontent.com/40576902/113406680-34893f80-93ac-11eb-9717-8a88dfa197f3.png)

As expected, it does not clear even a single byte of the cache (it couldn't have forgotten all of it, since I had cached and regularly closed Olive several times).

PR #1306, opened 2020-11-05
Add Clear Disk Cache to Sequence menu

Simran-B:
A faster way to clear disk cache while testing 0.2, see #1305

### DIFF
--- a/app/window/mainwindow/mainmenu.cpp
+++ b/app/window/mainwindow/mainmenu.cpp
@@ -22,6 +22,7 @@
 
 #include <QDesktopServices>
 #include <QEvent>
+#include <QMessageBox>
 #include <QStyleFactory>
 
 #include "common/timecodefunctions.h"
@@ -30,6 +31,7 @@
 #include "dialog/actionsearch/actionsearch.h"
 #include "dialog/task/task.h"
 #include "panel/panelmanager.h"
+#include "render/diskmanager.h"
 #include "tool/tool.h"
 #include "ui/style/style.h"
 #include "undo/undostack.h"
@@ -165,6 +167,8 @@ MainMenu::MainMenu(MainWindow *parent) :
   sequence_menu_ = new Menu(this, this, &MainMenu::SequenceMenuAboutToShow);
   sequence_cache_item_ = sequence_menu_->AddItem("seqcache", this, &MainMenu::SequenceCacheTriggered);
   sequence_cache_in_to_out_item_ = sequence_menu_->AddItem("seqcacheinout", this, &MainMenu::SequenceCacheInOutTriggered);
+  sequence_menu_->addSeparator();
+  sequence_cache_clear_ = sequence_menu_->AddItem("seqcacheclear", this, &MainMenu::SequenceCacheClearTriggered);
 
   //
   // WINDOW MENU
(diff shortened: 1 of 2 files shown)


TheTechRobo:
What's the status on this PR?

Simran-B:
Well, the intent of this PR is to make it easier to clear the cache. There is a problem however: Olive only deletes the files that it knows are cache files, but in case of a crash it may "forget" about some and thus not erase them. The likelihood of these being corrupted is high, and the can cause follow-up issues, see https://github.com/olive-editor/olive/issues/1284#issuecomment-729136248

So without further changes, this PR doesn't really improve alpha testing. A better cache clearing solution is needed.

elsandosgrande:
Could the cache files have something like an identifier in the file name? If, say, every cache file begins with `olive-editor-cache`, then it should be easy to avoid collisions with other files while being able to easily identify the file.

TheTechRobo:
@elsandosgrande and, on startup it picks a random number for the prefix, and anything that doesn't start with that prefix will be deleted?

elsandosgrande:
@TheTechRobo What do you mean by picking a random number for a prefix? Every cache file would begin with the string `olive-editor-cache` and thus be easily detectable within the designated directory.

TheTechRobo:
I was thinking because that way it wouldn't delete current cache files, but that's invalid since we'd be running that on startup.

So forget what I said. :cat_smile:

Simran-B:
It should clear the entire cache without accidentally deleting user files. There's no need to add a salt to make cached files from different sessions distinguishable.

ThomasWilshaw:
Would it maybe make more sense to give the cache exrs a custom extension like .ovexr or something?

sobotka:
> Would it maybe make more sense to give the cache exrs a custom extension like .ovexr or something?

Ideally there should be a point where external tools can pre-generate the cache files. Having the cache in a well documented and robust file format is an asset.

Simran-B:
What would be the benefit of allowing tools other than Olive to pre-generate the cache? Why not let Olive handle that, e.g. as a headless process that monitors a folder?

The problem is that Olive calculates a hash value for each frame which it uses to organize its mediacache. External tools would have to implement the same logic to place the files where Olive will look for them. The hash algorithm may change, which would break the pipeline. A possible trade-off would be to let external tools create EXRs and have Olive move/copy them into its cache file and folder structure. But it would still require additional information to let Olive know what the source is and its metadata (unless it's all embedded into each EXR).

I have mixed feeling about using a custom file extension. It makes it less clear what kind of file it is and double-click to open in standard app won't work unless we register `.ovexr`. A better approach might be to use `.olive-cache.exr` as extension. There's still the question whether the cache subfolders (`00`..`ff`) should be removed when clearing the cache, which complicates the cleanup procedure if we only want to remove folders which are empty after having deleted Olive's EXRs.

Another option would be to simply delete the entire mediacache folder and warn users if they change the cache path to an existing, non-empty folder. There would be no warning for the default cache path however.

elsandosgrande:
I would personally go with prefixing the file name with `olive-editor-cache` or `olive-cache`, but I completely agree that changing the final file extension should be avoided.

@Simran-B If the cache is already stored in specific sub-directories, as opposed to simply being stored in the cache directory without any sub-directories, then wouldn't it be simple enough to delete all of the EXR files with the marker string and then delete all of the sub-directories that are left empty after that operation?

Regarding external tools, I'm guessing that they could be used on ingestion stations to generate the initial cache while copying the footage over to the local network-attached-storage server(s) after transcoding if that is necessary. As I don't have any experience in this field, this is what I think of when I think of an ingestion station: [reference one](https://linustechtips.com/topic/1039341-what-are-ingest-stations/), [reference two](https://www.youtube.com/watch?v=eQED3tF8wuw)

PR #1315, opened 2020-11-08
Add support for transitions when loading OTIO files

ThomasWilshaw:
OTIO transitions are now added to the timeline as cross dissolves with
the appropriate in/out points,

Also added a safety check when dealing with media references.


elsandosgrande (approved):
This is working in my local builds (`tests/sample_data/clip_sample.otio` from the OpenTimelineIO repository does not result in any assertions being raised anymore), so I'd say that it's ready unless Thomas wants to add something.

elsandosgrande:
@ThomasWilshaw Would you say that this is complete? I found it to function well in my review.

ThomasWilshaw:
I think so yes

elsandosgrande:
@itsmattkc

PR #1329, opened 2020-11-14
OTIO Dialog

ThomasWilshaw:
OTIO files don't carry a project settings (frame rate. resolution etc.) so this add a dialog that allows the user to select the sequence settings before loading the file. 

Also adds ability for the OTIO loader to load files that don't exist and need relinking.  For example if a user is given an OTIO file and a harddrive, the file paths wont necissarily be correct.


ThomasWilshaw:
For somereason this PR isn't tracking branch changes so I'll reopen it.

PR #1330, opened 2020-11-15
OTIO load

ThomasWilshaw:
OTIO files don't carry a project settings (frame rate. resolution etc.) so this add a dialog that allows the user to select the sequence settings before loading the file


ThomasWilshaw:
I've converted this into a draft again as the file linking is still under development. Any comments are welcome though.

app/dialog/otioproperties/otiopropertiesdialog.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+namespace olive {
+
+OTIOPropertiesDialog::OTIOPropertiesDialog(QList<Sequence*> sequences, Project* active_project, QWidget* parent) :
+    QDialog(parent),
+    sequences_(sequences)
+{
+  QVBoxLayout* layout = new QVBoxLayout(this);
+
+  layout->addWidget(new QLabel("Change settings for each OTIO sequence to be loaded"));
```
itsmattkc, line 36: Missing `tr()` wrapper

app/dialog/otioproperties/otiopropertiesdialog.h
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * @brief Dialog to load setting for OTIO sequences.
+   *
+   * Takes a list of Sequences and allows the setting of options for each.
+   */
+  class OTIOPropertiesDialog : public QDialog {
+    Q_OBJECT
+   public:
+    OTIOPropertiesDialog(QList<Sequence*> sequences, Project* active_project, QWidget* parent = nullptr);
```
itsmattkc, line 23: Might be nice to change this to `const QList<Sequence*> &sequences` to try saving a deep copy.

app/core.cpp
```diff
@@ -379,6 +382,14 @@ void Core::DialogExportShow()
   }
 }
 
+#ifdef USE_OTIO
+void Core::DialogImportOTIOShow(QList<Sequence*> sequences) {
```
itsmattkc, line 386: Might be nice to change this to `const QList<Sequence*> &sequences` to try saving a deep copy (though Qt might deep copy here anyway when going between threads)

itsmattkc (changes requested):
Overall seems good, some minor things here and there, but once those are resolved I think this is probably good to merge

ThomasWilshaw:
I think that's all the issues resolved

app/dialog/otioproperties/otiopropertiesdialog.cpp
```diff
@@ -0,0 +1,80 @@
...
+namespace olive {
+
+OTIOPropertiesDialog::OTIOPropertiesDialog(const QList<Sequence*>& sequences, Project* active_project, QWidget* parent)
+    :
+    QDialog(parent),
+    sequences_(sequences)
+{
+  QVBoxLayout* layout = new QVBoxLayout(this);
+
+  layout->addWidget(new QLabel(tr("Change settings for each OTIO sequence to be loaded")));
```
itsmattkc, line 37: Prefer using "OpenTimelineIO" in user-facing strings rather than "OTIO". Also perhaps this could be a little more descriptive, e.g. mentioning OTIO files don't contain sequence parameters so please enter the correct parameters for each sequence here.

app/dialog/otioproperties/otiopropertiesdialog.cpp
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  layout->addWidget(table_);
+
+  QDialogButtonBox* buttons = new QDialogButtonBox(QDialogButtonBox::Ok|QDialogButtonBox::Cancel);
+  connect(buttons, &QDialogButtonBox::accepted, this, &OTIOPropertiesDialog::accept);
+  connect(buttons, &QDialogButtonBox::rejected, this, &OTIOPropertiesDialog::reject);
+  layout->addWidget(buttons);
+
+  setWindowTitle(tr("Load OTIO File"));
```
itsmattkc, line 69: Prefer using "OpenTimelineIO" in user-facing strings rather than "OTIO"

app/task/project/loadotio/loadotio.cpp
```diff
@@ -88,11 +89,36 @@ bool LoadOTIOTask::Run()
...
+  // Variables used for loading bar
+  float number_of_clips = 0;
+  float clips_done = 0;
+
+  // Generate a list of sequences with the same names as the timelines.
+  // Assumes each timeline has a unique name.
   foreach (auto timeline, timelines) {
-    // Create sequence
     Sequence* sequence = new Sequence();
     sequence->SetLabel(QString::fromStdString(timeline->name()));
```
itsmattkc, line 102: Might be good to check if `timeline->name()` is empty, and if so, use an auto-generated name like "Sequence 1", "Sequence 2", etc. Otherwise it potentially permits a sequence with an empty name which Olive code assumes will never happen.

app/task/project/loadotio/loadotio.cpp
```diff
@@ -88,11 +89,36 @@ bool LoadOTIOTask::Run()
...
+      auto otio_track = static_cast<OTIO::Track*>(track.value);
+      number_of_clips += otio_track->children().size();
+    }
+  }
+
+  // Dialog has to be called from the main thread so we pass the list of sequences here.
+  QMetaObject::invokeMethod(Core::instance(),
+                            "DialogImportOTIOShow",
+                            Qt::BlockingQueuedConnection,
+                            Q_ARG(QList<Sequence*>,timeline_sequnce_map.values()));
```
itsmattkc, line 118: Should return a bool from DialogImportOTIOShow if `QDialog::exec() == QDialog::accepted` and cancel loading here if false. I found clicking cancel on the dialog would still load the project, which feels a little odd.

itsmattkc (changes requested):
Sorry, just a few more minor usability things I noticed. After that, it should be all good.

app/task/project/loadotio/loadotio.cpp
```diff
@@ -88,11 +89,51 @@ bool LoadOTIOTask::Run()
...
+  int unnamedSequenceCount = 1;
   foreach (auto timeline, timelines) {
-    // Create sequence
     Sequence* sequence = new Sequence();
-    sequence->SetLabel(QString::fromStdString(timeline->name()));
+    if (!timeline->name().empty()) {
+      sequence->SetLabel(QString::fromStdString(timeline->name()));
+    } else {
+      // If the otio timeline does not provide a name, create a default one here
+      QString label = "Sequence " + QString::number(unnamedSequenceCount);
```
itsmattkc, line 107: Should use `tr("Sequence %1").arg(unnamedSequenceCount)` so the string can be translated correctly. Also, code style requires underscored variables rather than camel-case (i.e. `unnamed_sequence_count`)

ThomasWilshaw: Sorry, can't believe I just did that (missed the `tr()` that is)

PR #1335, opened 2020-11-17
Bring back old translations and update them

prokoudine:
Old translation are automatically marked as fuzzy, so they only need to be confirmed where applicable. This should be a good start instead of forcing people to do it all over again :)

### DIFF
--- a/app/ts/CMakeLists.txt
+++ b/app/ts/CMakeLists.txt
@@ -15,6 +15,21 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 set(OLIVE_TS_FILES
+  ts/ar_AR.ts
+  ts/bs_BS.ts
+  ts/cs_CS.ts
+  ts/de_DE.ts
   ts/en_US.ts
+  ts/es_ES.ts
+  ts/fr_FR.ts
+  ts/id_ID.ts
+  ts/it_IT.ts
+  ts/pt_BR.ts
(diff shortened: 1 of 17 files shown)


Simran-B:
Wouldn't it be better to keep obsolete translations until someone checked whether they can be matched up with new translatable strings?

For example, "Project Save As..." was apparently changed to "Project Save As". The former had a translation in 0.1 in the file `ar_AR.ts` ( أحفظ المشروع ك...) but this PR only has the source string and no translation. It would only be a matter of removing `...`, for which you don't even need to know the language.

prokoudine:
I can't confirm that. The translation _is_ present here. Anyway, this is not how lupdate works. Before marking something as obsolete, it tries to find closest looking matches, inserts them, and marks as fuzzy. Then it removes obsoletes.

image

elsandosgrande:
I was about to say that looking at the `.ts` files alone probably doesn't paint the whole picture. I'll see how things stand with my translations and Qt Linguist and report back, but this pull request seems fine to me for now.

Simran-B:
My bad, it's `Save Project &amp;As` now. That didn't show up in my quick search. Should be fine fine.

elsandosgrande:
As far as Bosnian and Serbian go, everything's fine.

elsandosgrande (approved):

Simran-B (approved):

PR #1393, opened 2020-11-21
Update cs_CS.ts

pafri:
Update of Czech localization file. Note: the usual code for the Czech language in the filename, if a pattern such as "olive_cs.ts" is not used, is: cs_CZ.


Simran-B:
You're right, the file should be renamed to `cs_CZ.ts`.

elsandosgrande:
@pafri The filename still seems to be `cs_CS.ts`, so I'm guessing that you haven't yet renamed it. You can do so and add the commit to this pull request if you want.

pafri:
I suggest to leave the possible renaming of the file on developers. My contribution simply consists of using the GitHubs "edit tool" offered by GUI, copy and paste the content of the .ts file opened in text editor (after the finishing of the translation work in QtLinguist editor).

elsandosgrande:
Would it not have been simpler to just clone the repository and directly edit the file in Qt Linguist? Anyways, @pafri, I think that those who work on translations should manage them in their entirety, including the file names, but if you don't want to correct the name of the file, I'll just do it once this is merged. Speaking of which, is this translation complete, or is there some work left to do? I see over eight-hundred changed lines, which is roughly the total number of translatable strings, at least according to `lupdate` (859 lines if I recall correctly).

Simran-B:
@pafri you can rename the file directly on GitHub:

Screenshot_20201121-214310~2
Screenshot_20201121-214342~2

pafri:
Thanks for committing the file. The only way I am able to change the filename, is to do it in file manager. Also, as I wrote, I used QtLinguist (as specialized editor) after downloading the file to edit (cs_CS). After finishing the job (yes 100 %), I used the "pull request" tool after replacing the content of the file in Git by the content of the changed file. It's quite simple way, how to do the contribution in our case. The same could be repeated (when currently the localization of the project Olive isn't managed by translation server like Transifex, Weblate, ... etc.), when there would be the need (when I see, that the number of fuzzy or untranslated strings is high enough (reasonable) for me to make new completion.

elsandosgrande:
@pafri You forgot to [update the file's `app/ts/CMakeLists.txt` entry](https://github.com/olive-editor/olive/wiki/Translations#building-your-translation-into-olive). I already pushed the update.

Out of curiosity, why did you download the file from GitHub instead of just cloning your own GitHub fork locally, modifying the file (renaming it, adding missing translations, …), modifying `app/ts/CMakeLists.txt` accordingly, pushing the commits to your fork on GitHub, and then creating a pull request? I mean, downloading a single file manually is uncommon in my, admittedly limited, experience at least, so I'm curious as to what circumstances led you to going this route. I hope that I'm not being a bother :sweat_smile:.

PR #1398, opened 2020-11-22
Translation fixes for cs_CZ

mcDandy:
Fixed phrasing of dialogs and some nonintuitive texts

### DIFF
--- a/app/ts/cs_CZ.ts
+++ b/app/ts/cs_CZ.ts
@@ -301,11 +301,11 @@ Délka: %4</translation>
     <name>main</name>
     <message>
         <source>Show this help text</source>
-        <translation>Ukázat tento text s nápovědou</translation>
+        <translation>Ukáže tento text s nápovědou</translation>
     </message>
     <message>
         <source>Show application version</source>
-        <translation>Ukázat verzi programu</translation>
+        <translation>Ukáže verzi programu</translation>
     </message>
     <message>
         <source>Start in full-screen mode</source>
@@ -313,7 +313,7 @@ Délka: %4</translation>
     </message>
     <message>
         <source>Export only (No GUI)</source>
-        <translation>Pouze exportovat (žádné rozhraní)</translation>
+        <translation>Pouze exportovat (bez rozhraní)</translation>
     </message>
     <message>
         <source>Override language with file</source>
@@ -532,15 +532,15 @@ Délka: %4</translation>
     <name>olive::Core</name>
     <message>
         <source>Import error</source>
-        <translation>Chyba při zavádění</translation>
+        <translation>Chyba při importu</translation>
     </message>
     <message>
         <source>Nothing to import</source>
-        <translation>Nic k zavedení</translation>
+        <translation>Nic k importu</translation>
     </message>
     <message>
         <source>Importing...</source>
-        <translation>Zavádí se...</translation>
+        <translation>Importuje se...</translation>
     </message>
     <message>
         <source>Import footage...</source>
@@ -648,7 +648,7 @@ Délka: %4</translation>
     </message>
     <message>
         <source>This Sequence is empty. There is nothing to export.</source>
-        <translation>Tento úryvek je prázdný. Není co vyvádět.</translation>
+        <translation>Tento úryvek je prázdný. Není co exportovat.</translation>
     </message>
     <message>
         <source>No valid sequence detected.
@@ -720,7 +720,7 @@ Ujistěte se, že je úryvek načten a má připojený uzel prohlížeče.</tran
     </message>
     <message>
         <source>Failed to cache sequence</source>
-        <translation>Úryvek se nepodařilo uložit do vyrovnávací paměti</translation>
+        <translation>Nepodařilo se uložit úryvek do vyrovnávací paměti</translation>
     </message>
     <message>
         <source>No active viewer found with this sequence.</source>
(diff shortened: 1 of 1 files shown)


elsandosgrande:
@mcDandy Could you provide a description please?

mcDandy:
Edited PR to contain description. 
Should have made it a draft, becoue there is still something left ("export" -> "vyvést", "import" -> "zavést", "ortographic matrix" -> "pravopisná matice"), but I can easily make another whenever I will have time.

elsandosgrande:
@mcDandy I'll make it a draft then. Just say whenever it's done in the future. Also, since one of the strings in the source code was changed (https://github.com/olive-editor/olive/commit/e70e70ad98e825e129cf138f3550e1800fa4013a), you'll need to update the translation to account for that as well.

mcDandy:
There are still strings, that does not feel right (hit you in the face), but I am not sure how to translate them and do not want to sit on the changes for nothing.

pafri:
Also finally we reached the next step and the changes  could be committed.
Note: the changes made during last week(s) in the previous version of the localization file by this draft are further reflected in the current branch of that file (some in this draft seen changes are (at least in some form) re-used, some not) and are as localization "alternative" available for download here: https://obvyklemisto.webgarden.cz/rubriky/soubory-ke-stazeni/preklady-rozhrani
Best regards.

elsandosgrande:
@mcDandy Could you mention a few of the in-your-face lines of text? I'd like to watch out for them when translating to the Serbo-Croatian languages :wink:.

mcDandy:
I have troubles with these. They are not wrong, but not right.
1 footage -> "záznam" I would translate it most likely to "klip" (not sure)
2 "No Active Sequence" -> "Žádný činný úryvek (sled záběrů)" 
3 "Set node label" -> "Nastavit štítek uzlu"
4 "Zoom to Fit" -> "Zvětšit pro přizpůsobení"
5"Performance"-> "Provedení"
6 "Failed to set color configuration: %1" "Nepodařilo se nastavit nastavení barev: %1"
7 "Edit Tool Also Seeks" "Nástroj pro úpravy také přehrává normální rychlostí s přeskakováním"
8 "Renderer" -> "Vykreslovač"
9 Solid -> Plný (The node name)

Simran-B:
> 1 footage -> "záznam" I would translate it most likely to "klip" (not sure)

Assuming that "klip" means clip, it would be an incorrect translation. Footage are file based media, auch as an MP4 or WAV file. The footage node represents the load operations.

Clips are containers for footage. They define which frames/samples to take from the media by defining an in and out point each, and define where on the timeline that piece of footage is to be played.

elsandosgrande:
I'd personally go with “snimak/снимак”, which translates back to “recording” (I'm guessing that there is a cognate in Czech). @Simran-B Would you agree with this approach?

Simran-B:
záznam apparently means recording too, so not sure. Footage isn't the best word perhaps, because it primarily means image material, which doesn't really cover audio. Raw material or media work better, although the latter can be too broad (include things that are not file based). Maybe prostředek (medium, means) is a good alternative?

elsandosgrande:
I get anything from “means”, as in “This word means…”, to “resource” when trying to translate “prostředek” into Bosnian, Serbian, and Croatian in Google Translate and Bing Translator, so maybe it's not the best pick? By the way, both translation engines(?) agree with the footage→recording approximation, though that does leave images out (according to the definition given by Google Translate, so does footage though).

pafri:
To "footage" translated as "záznam", meaning "recording": it is universal, although we might assume that (the word "footage") it is reserved for (only) video footage; also eg a (television) "recording" (= (televizní) záznam) is certainly not just a silent film, without audio/sound/noise track; note: there seems to be a certain shift in meaning during last decades (?), but / and nowadays we also just say "I'm watching video" too, so expressly (verbatim) not "I'm watching video and audio".

elsandosgrande:
@Simran-B, so, what's your verdict? @itsmattkc, I think that you should comment on this as well, given that you're the core developer of Olive, which means that you know exactly which meaning you intended to convey when choosing the wording which you chose when writing the strings in question.

Simran-B:
I'm still undecided. Footage seems too narrow and media too broad. File based media is too long.

What do nodes actually represent? Entities or actions? E.g. is it "a filter" or "to filter"? Does a footage node represent the actual file or does it stand for the task of reading from a file? Nuke has Load and Save nodes AFAIK. But overall, the node graph metaphor seems to work better if we think of nodes as entities and the edges as data flow. That would mean that we can't go with a simple "load" but need to decide whether it should be called footage, media or something else. It could just be called file, but that might not match up with the underlying implementation if there's a hybrid node for audio/video and separate nodes for other media types.

elsandosgrande:
@Simran-B In my head, nodes perform operations and the connections pass the results along. Perhaps we could consider footage/recordings to be anything recorded by a camera, whether that be still or in motion, with visuals or without visuals (pure audio)?

 Maybe we should ask @sobotka for his thoughts on the matter as well?

itsmattkc:
What's the status of this PR at the moment?

TheTechRobo:
@itsmattkc I believe they're deciding on the best translations and asking for some input from you.

Simran-B:
@itsmattkc The most discussed point above would be settled by changing Olive's terminology from _footage_ to _media_ for input files (we talked about that on Discord and you were to open to rename it).

itsmattkc:
Indeed, if "media" is preferable, I think it would be fine to change it to that. Particularly since now the "media node" and "footage item" have now become the same object.

mcDandy:
The status is that this file is very outdated agnist english file. I did not have time (and will not probably have much time for rest of semester (untill june)) also forgot about it.

mcDandy:
I do not know. We should do with path of least resistence (easyer).

Simran-B:
Merging this for now and will try to resolve conflicts in the other PR.

Issue #1331, opened 2020-11-16
[UI] Node Editor toolbar (proposal)


pgilfernandez:
**Commit Hash** 
2d2d6d2a

**Platform** 
All

**Summary**
In order to improve the user experience I would like to suggest adding a toolbar to the Node Editor, something like the following mockup:

<img width="555" alt="nodes_icons_v01" src="https://user-images.githubusercontent.com/5942369/99239454-60d86100-27fb-11eb-98d9-683dc05daf1a.png">

- the "add" icon would be aligned to the left as it is a real node tool
- the other icons are aligned to the right as they all are just visibility options
- "smooth edges" has 2 states: activated, deactivated
- "filter" has 2 states: activated, deactivated
- ony 1 of the 4 "direction" icon activates at a time

I have already designed the icons in SVG so, if you like the idea, I would straightforward PR them all.

**Additional Information / Output**

GUY-BBS:
Hi Matt, While I appreciate that your top priority is getting version 2 stable, could I request you eventually consider a copy/save feature in Fernandez  proposed tool bar to store and recover useful node configurations and make them available to other projects?

pgilfernandez:
@GUY-BBS would you like an import/export set of icons or a copy/paste ones?
I can design both if needed ;)

GUY-BBS:
@pgilfernandez  Ideally both. Initially, it takes some time to impliment something like an animated name strap using nodes in Olive. It would be useful to copy and paste it within the same project and it would be great to save to a library of similar effects for use in other projects.

sobotka:
Node bundling is a large design task. If rushed, it will bring pain.

pgilfernandez:
On my side it's no problem to design and PR those 4 icons, but I guess the implementation would need much more effort...
Anyway, the proposed toolbar just uses already working features so it may be easy to code... it would be a start point.

pgilfernandez:
Here is a new mockup with the icons already designed in SVG:

<img width="555" alt="nodes_icons_v02" src="https://user-images.githubusercontent.com/5942369/99827503-0112ea80-2b5a-11eb-9e46-668f119a26aa.png">

I would suggest to add some spacing between groups of icons, but I don't know if a "separator" is already available in the project, something like the following:

<img width="555" alt="nodes_icons_v03" src="https://user-images.githubusercontent.com/5942369/99828030-b47bdf00-2b5a-11eb-8056-4b21890e179c.png">

[...]

PR #1399, opened 2020-11-22
[UI] Nodes View toolbar iconset

pgilfernandez:
As these icons don't interfere with more important tasks I'm PR them so that they are already available when needed.
As commented in [in issue #1331](https://github.com/olive-editor/olive/issues/1331) the proposed toolbar has a "phase 1" that could already be implemented and a future "phase 2":
PHASE 1:
- smooth edges
- filter
- nodes direction (4 options)

PHASE 2:
- copy/paste
- import/export


elsandosgrande:
@pgilfernandez Is this intended to be merged as-is, or will there be further work in the future? I mean, the impression which I get from reading the accompanying ticket is that this is still at a draft stage, in which case I could convert the pull request to a draft.

pgilfernandez:
@elsandosgrande they are just the icons. I'm not an experienced coder so I can just help in the graphic design side. A coder will be needed to actually create the toolbar, add the buttons with the icons and add the functionality.
What I call "phase 1" should be pretty easy for a coder because it's just a front end interface and link the buttons to the commands that are already implemented and that are accessible via right mouse button menu.

"Phase 2" would need further work, not easy at all... but, well, I have added the icons because they are pretty common and they could be used anywhere else in the software.

elsandosgrande:
@pgilfernandez :thinking: 

I don't think that it's all that desired to have unreferenced assets in the main repository, so maybe this could be merged into a feature branch with the “second phase” being merged there as well later down the line? @itsmattkc, what are your two cents?

PR #1402, opened 2020-11-23
Update zh_CN.ts

XGSZZ:
Continued translating the zh_CN.ts file, and modified the translation of several strings.


elsandosgrande (approved):
There are no conflicts or errors with local builds, so it's 110% ready to be merged.

PR #1414, opened 2020-11-24
openglrenderer: fix destroying surface assert error

ThomasWilshaw:
DestroyInternal() was trying to destroy a surface that was in a
different thread leading to a sendEvent Assert error. This is fixed
by invoking the destroy() method in the correct thread.

### DIFF
--- a/app/render/opengl/openglrenderer.cpp
+++ b/app/render/opengl/openglrenderer.cpp
@@ -145,7 +145,7 @@ void OpenGLRenderer::DestroyInternal()
 
     // Destroy surface if we created it
     if (surface_.isValid()) {
-      surface_.destroy();
+      QMetaObject::invokeMethod(surface_.thread(), "destroy", Qt::AutoConnection);
     }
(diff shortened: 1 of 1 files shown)


itsmattkc:
Thanks for the PR but I don't think this is correct. While you're right that the surface shouldn't be destroyed in another thread, you shouldn't send the signal to `surface_.thread()`, that's effectively just sending it to a `QThread` object rather than the object itself.

Secondly, `destroy` isn't a slot on the `QOffscreenSurface` class, which I believe is necessary for `invokeMethod` to work. As such I believe this effectively amounts to a no-op, which indeed would address this since it isn't try to destroy the surface in another thread, but I think it's not trying to destroy it at all.

ThomasWilshaw:
I thought that might be the case after your earlier comment. No problem I'll close this then

itsmattkc:
Thanks for calling attention to it though, I can write a quick fix for it now

PR #1416, opened 2020-11-25
Improved translation

XGSZZ:
Continued translating zh_CN.ts file.
Modified several Chinese characters.


ShuiHuo:
@XGSZZ Can use Chinese punctuation?

PR #1422, opened 2020-11-27
Use appropriate labels for media nodes

ZoomTen:
Set imported Media node label to "Video Input" for video footage, and "Audio Input" for audio footage

This was prompted by troubleshooting a friend's node setup, and having unlabelled Media nodes made things kind of confusing (e.g. they would accidentally link a Volume node to a Media node containing only video)

### DIFF
--- a/app/core.cpp
+++ b/app/core.cpp
@@ -1081,7 +1081,7 @@ void Core::LabelNodes(const QVector<Node *> &nodes) const
 
   if (ok) {
     foreach (Node* n, nodes) {
-      n->SetLabel(s);
+      n->SetLabel(s, true);
     }
   }
 }
--- a/app/node/input/media/media.cpp
+++ b/app/node/input/media/media.cpp
@@ -23,6 +23,8 @@
 #include "common/timecodefunctions.h"
 #include "common/tohex.h"
 
+#include "project/item/footage/stream.h"
+
 namespace olive {
 
 MediaInput::MediaInput() :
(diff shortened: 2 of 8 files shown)


Simran-B:
Looks like Alex accidentally referenced the wrong issue (1422 instead of 1421). This PR looks good to me.

ThomasWilshaw:
Might it be better to add a function to the media node that sets the label based on the stream type? That way if you add node through the UI and select say an audio stream the label becomes "Audio" as well?

ZoomTen:
@ThomasWilshaw 

You have a point there, so I just pushed some additional commits that do just that, though my approach may be flawed

ThomasWilshaw:
I think you should be able to simplify this a lot and keep it all in `media.cpp`. You shouldn't need the `HasUserLabel` method, just grab the current label in `FootageChanged` and compare it to what it should be autoamticaly. You'll probalably also need to call `FootageChanged` in `SetStream` . Finally you'll need to update `Retranslate` and add some logic there.

itsmattkc:
Thanks for the PR, but I don't know if this exactly solves the problem. Before we merged the video and audio nodes, this is how they were named, and there was still confusion surrounding what exactly "Video Input" and "Audio Input" meant.

For example, if your node graph looked like "Video Input" -> "HelloWorld.MP4 (Clip)", people were confused about which node was the footage and which one was the clip (often they assumed the clip was the footage because it was labelled with the footage filename).

TL;DR, I think this is a much deeper issue than node labelling and requires a different approach.

PR #1423, opened 2020-11-27
Russian translation - Fix one instance of plural forms

prokoudine:
Partially fixes #1421. Not sure about other instances yet.

PR #1424, opened 2020-11-27
Fix plural handling for translations

### LINKED ISSUE #1421: [TRANSLATION] [UI] Missing plural forms support
Here are all the instances (that I found) of translatable text that needs plural forms support:

./app/project/item/footage/audiostream.cpp: return QCoreApplication::translate("Stream", "%1: Audio - %2 Channels, %3Hz").arg(QString::number(index()),
./app/widget/nodeparamview/nodeparamviewarraywidget.cpp: count_lbl_->setText(tr("%1 elements").arg(array_->GetSize()));
./app/dialog/preferences/tabs/preferencesdisktab.cpp:  cache_ahead_slider_->SetFormat(tr("%1 seconds"));
./app/dialog/preferences/tabs/preferencesdisktab.cpp:  cache_behind_slider_->SetFormat(tr("%1 seconds"));
./app/dialog/preferences/tabs/preferencesgeneraltab.cpp:  default_still_length_->SetFormat(tr("%1 seconds"));
./app/task/project/import/import.cpp:  SetTitle(tr("Importing %1 files").arg(file_count_));

Simran-B:
https://github.com/olive-editor/olive/commit/992b6eb95217554dc309cf85686ac9dddab4cb8b does not fix #1421. The placeholder `%n` must be used instead of `%1`, `%2` etc. and the value passed to `tr()` instead of via `arg()`.

In the current state, this PR breaks widgets such as the FloatSlider. When entering a new value or dragging it, the label does not update. There are also QString::arg warnings, because a former placeholder such as `%1` no longer exists (replaced by `%n`). It will require some refactoring to properly update the UI I'm afraid, but I couldn't figure out how so far.

This PR also replaces the English translation file with `lupdate . -ts app/ts/en_US.ts -pluralonly`, which only contains the strings which require plural handling. I added translations for the few strings that do. As long as the source strings are in American English, then we don't need a full translation (unless we want to allow users to change the English UI?!)

TODO:
- [ ] Update Translation Wiki
- [ ] Change build system so that CMake updates `en_US.ts` with `-pluralonly` option and document this special case in Wiki.
       -> The logic to conditionally create or add translations got removed here:
            https://github.com/olive-editor/olive/commit/b9724e391074e6c78c4319a6e511df6d6f22ce76#diff-fd99ceeacdfe207bd1168e1c30c94ed9f316895e74154c83258807c703288b3fL272-L276
             `qt5_add_translation` is called unconditionally for updating .qm files https://github.com/olive-editor/olive/blob/c03c19741254f0fecfbf7dab2515a2da0a297ce7/app/CMakeLists.txt#L63
             The `UPDATE_TS` option does nothing and there is no call to `qt5_create_translation` anywhere that would update .ts files. So there isn't anything to change to pass `-pluralonly`...

Obsoletes #1423


prokoudine:
I guess that obsoletes my patch in https://github.com/olive-editor/olive/pull/1423 :)

itsmattkc:
Thanks for picking this up @Simran-B and @prokoudine, I misunderstood how Qt's plurals worked.

You may have noticed this already, but this might get complicated with those `SetFormat` calls in the Preferences dialog. The sliders use the format as a means of printing the value, while you can override the format and place the value directly in it, the slider won't know how to update it automatically anymore.

Simran-B:
Yeah, it's certainly not trivial. One could set the string in SetFormat() without calling tr() and do that dynamically sometime later. That might make the UI work, but I'm fairly sure that lupdate won't pick up the source string anymore. Pseudo code:

```
setFormat("%n foo");
...
tr(getFormat(), nullptr, Value().toInt()); // <-- not a string literal
```

Maybe an unused/unreachable piece of code could be added with the string literal so that it gets picked up by lupdate without interfering with get/setFormat?

The easiest solution would be to not add plural handling for sliders and such however.

itsmattkc:
Yeah pretty sure lupdate won't pick up `tr()` without an immediate string.

A hacky solution might be to connect a function to the sliders' "value changed" signal and update the format externally each time it's value changes? That's probably the simplest solution that actually makes it work, but it's definitely hacky.

Otherwise yeah I'd just be inclined to not handle plurals for those.

app/dialog/preferences/tabs/preferencesgeneraltab.cpp
```diff
@@ -91,10 +92,11 @@ PreferencesGeneralTab::PreferencesGeneralTab()
 
   general_layout->addWidget(new QLabel(tr("Default Still Image Length:")), row, 0);
 
+  value = Config::Current()["DefaultStillLength"].value<rational>().toDouble();
   default_still_length_ = new FloatSlider();
   default_still_length_->SetMinimum(0.1);
-  default_still_length_->SetFormat(tr("%1 second(s)"));
-  default_still_length_->SetValue(Config::Current()["DefaultStillLength"].value<rational>().toDouble());
+  default_still_length_->SetFormat(tr("%n second(s)", nullptr, value));
+  default_still_length_->SetValue(value);
```
Simran-B, line 99: This is incomplete

Issue #1175, opened 2020-09-13
❗❗ READ BEFORE REPORTING AN ISSUE ❗❗


itsmattkc:
First of all, thank you for your interest in Olive. Olive is currently undergoing a large-scale transition from 0.1 to 0.2, and to make the most of our limited resources, we are only accepting certain issue reports at this time.

**Our current focus is on making Olive 0.2 a stable and usable program for editing video, so we would appreciate if you primarily provide issues that support this goal.**

### ❌ 0.1 Issues Are Unsupported
We are no longer supporting 0.1 (aka "April 2019 | 1e3cf53") and any bug reports or feature requests pertaining to it will be closed. All of our development focus is on 0.2, so if you're able to, start testing 0.2 so we can get it stable and usable sooner!

### ✔️ Crashes
If you're using 0.2 and run into any fatal errors or terminations, we would love to know more. Please provide as much detail about how the crash was triggered as possible and if you can recreate it, provide a debugger backtrace ([consult this wiki article if you don't know how](https://github.com/olive-editor/olive/wiki/Debugging-Olive)).

### ❌ Most Feature Requests Will Be Rejected
In this heavily transitional time, the focus is on making Olive a functional and stable program with a consistent and usable design. Therefore, with very few exceptions, we are not accepting feature requests at this time as they frequently distract from this goal. If you're uncertain, perhaps consult the Discord server before making a request.

### ✔️ Bugs or behavioral issues (particularly caching, rendering, color management, and nodes)
If you're using 0.2 and run into any issues related to the editing process (particularly when it comes to caching/previewing), please let us know. The preview cache is very much intended to be reliable and robust while also being performant, which can be a difficult balance to strike and we need your help to get there.

Thank you for your co-operation and hopefully 0.2 will be in a ready state soon.

PR #1426, opened 2020-11-28
new glitch efffect!

jgarza9788:


Simran-B:
Olive 0.1 is not supported anymore: #1175

jgarza9788:
Thanks, I’ll add it to the recent version of olive and re-submit!

PR #1433, opened 2020-11-30
Update zh_CN translation

XGSZZ:
Continued translating zh_CN.ts file.
Changed two sets of quotation marks to English, I not sure whether quotes should be change to Chinese punctuation.


elsandosgrande:
@XGSZZ It's your choice at the end of the day, but I vote for Chinese punctuation.

XGSZZ:
Changed to Chinese punctuation.

PR #1434, opened 2020-11-30
cmake: misc macOS fixes

eszlari:
(not tested, because I have no Apple devices)


itsmattkc:
What's the point of removing `OLIVE_TARGET`?

Simran-B:
Also, why remove `set(CMAKE_OSX_DEPLOYMENT_TARGET)` from CMakeLists.txt and set it in the CI only? It seems like a meaningful default to me and since set() isn't FORCEd, it can already be overruled when configuring.

eszlari:
> What's the point of removing `OLIVE_TARGET`?

When I wrote the cmake build system, I didn't know about `OUTPUT_NAME`. The plan is to get ride of superfluous variables:

`OLIVE_TARGET` > [set_target_properties(OUTPUT_NAME ...)](https://cmake.org/cmake/help/latest/command/set_target_properties.html)
`OLIVE_DEFINITIONS` > [target_compile_definitions()](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html)
`OLIVE_SOURCES` & `OLIVE_RESOURCES` > [target_sources()](https://cmake.org/cmake/help/latest/command/target_sources.html)

https://github.com/boostcon/cppnow_presentations_2017/blob/master/05-19-2017_friday/effective_cmake__daniel_pfeifer__cppnow_05-19-2017.pdf (p. 14)


> Also, why remove set(CMAKE_OSX_DEPLOYMENT_TARGET) from CMakeLists.txt and set it in the CI only? It seems like a meaningful default to me and since set() isn't FORCEd, it can already be overruled when configuring.

At least it should be moved before `project()`:

https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html

> The value of this variable should be set prior to the first project() or enable_language() command invocation because it may influence configuration of the toolchain and flags. It is intended to be set locally by the user creating a build tree. This variable should be set as a CACHE entry (or else CMake may remove it while initializing a cache entry of the same name).

itsmattkc:
> When I wrote the cmake build system, I didn't know about OUTPUT_NAME. The plan is to get rid of superfluous variables:

Yeah, but we use a different target name on macOS... which makes it no longer superfluous, right?

eszlari:
> Yeah, but we use a different target name on macOS... which makes it no longer superfluous, right?

A target is just build system concept. In cmake by default, the file name of the executable is inferred from the target name. By setting `OUTPUT_NAME`, you can chose a different name for the executable.

itsmattkc:
Gotcha. Sorry for the late response on this, but looks all good.

PR #1444, opened 2020-12-14
ValueNode

### LINKED ISSUE #1443: [NODES] String / Number node
**Commit Hash** 
1cd07bc1

**Platform** 
macOS 10.14.6

**Summary**
I think it has been already said somewhere but I can't find a issue/feature request that covers it so, let me write it down in case you think it's important:

I'm missing a simple node that inputs a number (integer, float,...) or a simple node that inputs a text. Then they can be connected to any value of other nodes so that you save time and have everything "parametric".
For instance, I have done this little node group to stroke and drop a shadow of a text but, by now, I'm force to write the text in both nodes. I have also added in red a mockup of the "string node" that could be connected to the "text parameter" inside the Text node:
string

The same would apply for numbers and any numeric value inside any node.

Thanks

[EDIT]: Additionally I guess there are another kind of "input nodes" such as a boolean, percentage, 2D vector (X and Y numeric values), 2D vector with percentage, ...

ThomasWilshaw:
Initial implementation of a node that generates a value, I'd like to extend this to be able to input vectors and so on but I don't know how possible that will be.

Fixes #1443


ThomasWilshaw:
This is being put on hold for now until the node system has had some updates/redesisgns

PR #1446, opened 2020-12-17
Rational slider

ThomasWilshaw:
Adds a rational slider that can display values as either timecode, timestamp, rational or float. At the moment Ctrl clicking the slider changes the display type. As this is still a draft I've added the slider to the sequence dialoge for testing/experimenting. Some questions/thoughts:

1. As this is mainly for displaying timecodes I've set the minimum increment to be the timebase of the slider, is this reasonable?
2. Is there a good way to make sure sliders always pick up a change in timebase and should the timebase be global or local to the slider? If the slider is displaying the in point for a clip for example, should that be in the clip's native timecode or the sequence's global timecode?
3. is Ctrl clicking a good way to change displays?
4. I feel like ladders in this case don't really make sense unless we implement a custom ladder (frames->seconds->minutes->etc.).

I will squash all these commis down when it's ready to be merged.


sobotka:
Custom ladder is a very interesting idea. I wonder if that can be fleshed out further into a class for more useful display?

Is it plausible to distill the information down to make the ladder parametric? Perhaps a virtual function for the formatting?

ThomasWilshaw:
The slidder ladders are alread a class of their own so it would be fairly straightforward to make a custom one (I think...). What sort of parameters are you thinking? At the moment the ladders are based on a multiplier, but oyu could possibly pull of the slider type (`kInt, kRational` etc.) to get customisation.

sobotka:
The two primary parametric driven components are likely formatting and base / central increment? Thoughts?

ThomasWilshaw:
The formatting is already handled by the slider itself (so the ladder on the rational slider displays 00:00:00:00 or 0/0 etc.). I don't think base would really make sense in this case as although the slider is technically for rationals it has mainly been written with timecode in mind so there won't be a constant incremant.

sobotka:
That’s what I mean; use a virtual function for the formatting so it can be customized to more contexts. For the increment, perhaps setting a log or uniform action? Define the basic five increments? Does that make sense?

ThomasWilshaw:
Note: I can probably update a lot of the `Set...()` code to use `QVariant::fromValue()`

TheTechRobo:
Failing on GCC and Clang (macOS = Clang; Linux = GCC/Clang)

app/widget/slider/sliderlabel.cpp
```diff
@@ -54,9 +54,11 @@ void SliderLabel::mousePressEvent(QMouseEvent *e)
   if (e->button() == Qt::LeftButton) {
     if (e->modifiers() & Qt::AltModifier) {
       emit RequestReset();
-    } else {
+    } else  {
       emit LabelPressed();
     }
+  } else if (e->button() == Qt::RightButton) {
```
itsmattkc, line 60: Rather than catching `Qt::RightButton`, it would be better to set `setContextMenuPolicy(Qt::CustomContextMenu)` on the label. Then the label will automatically emit a `customContextMenuRequested` signal that you can connect to instead of a new signal.

app/widget/slider/sliderbase.cpp
```diff
@@ -401,4 +450,11 @@ void SliderBase::RepositionLadder()
   }
 }
 
+void SliderBase::ChangeSliderType()
+{
+  if (mode_ == kRational) {
+    emit changeRationalDisplayType();
```
itsmattkc, line 456: Would prefer the RationalSlider do this itself than have to go through SliderBase, but you might need to add a `protected` accessor so that RationalSlider can get it (something like `QLabel *label() { return label_; }`)

app/widget/slider/sliderladder.cpp
```diff
@@ -291,6 +291,12 @@ void SliderLadderElement::UpdateLabel()
   } else {
     label_->setText(value_);
   }
+
+  // Increase the ladder width if displaying timecode
+  if (value_.contains(":")) {
+    label_->setFixedWidth(QtUtils::QFontMetricsWidth(label_->fontMetrics(), QStringLiteral("00:00:00:00")));
```
itsmattkc, line 297: Seems a little hacky to check the string for a colon. It might be better for the slider itself to provide the minimum width on the SliderLadder's construction, which can then be set on SliderLadderElement's construction.

itsmattkc (changes requested):
Apart from a few notes, it seems all good! (though I am struck by some issues in the original SliderBase code that I think I'll want to fix at some point, but I'll do that post-merge so it doesn't interfere)

itsmattkc:
Looks good, let me know when it's ready to merge and I'll go ahead and do that.

ThomasWilshaw:
I think is all good to go now.

PR #1477, opened 2021-01-30
cmath library needed for std::log10

morrolinux:


itsmattkc:
Ahh I didn't see this PR, I just pushed this fix to the branch

PR #1489, opened 2021-02-18
Docker: Update build image with final OCIOv2

Simran-B:
- Update to OpenColorIO v2.0.0 release version
- Update OpenTimeIO build and adjust to their new build system
  (might require additional changes to fully support --prefix)
- Use new ASWF's new qtbase package, CY2021 now actually being the correct VFX platform version (Qt v5.15.2)
- Add missing `xcb-util-*` libraries which seem to be required now with the updated Qt
- Cleanup


cmake/FindOpenColorIO.cmake
```diff
@@ -51,7 +51,6 @@ find_library(OCIO_LIBRARY
         libOpenColorIO.so.2.0 # libOpenColorIO.so.2.0 (Linux)
         OpenColorIO.2.0       # libOpenColorIO.2.0.dylib (macOS)
         OpenColorIO_2_0       # OpenColorIO_2_0.lib (Windows)
-        OpenColorIO           # (fallback)
```
Simran-B, line 54: This seems to still be required for the Windows deps package https://olivevideoeditor.org/deps/dep-win.7z (IIRC, OpenColorIO_2_0 is renamed to just OpenColorIO there)
https://github.com/olive-editor/olive/pull/1489/checks?check_run_id=1929778172

docker/ci-otio/Dockerfile
```diff
@@ -9,9 +9,10 @@ ARG ASWF_PKG_ORG=aswftesting
 ARG CI_COMMON_VERSION=2
 ARG VFXPLATFORM_VERSION=2021
 
-FROM ${ASWF_PKG_ORG}/ci-package-python:${VFXPLATFORM_VERSION} as ci-package-python
-FROM ${ASWF_PKG_ORG}/ci-package-qt:${VFXPLATFORM_VERSION} as ci-package-qt
-FROM ${ASWF_PKG_ORG}/ci-package-pyside:${VFXPLATFORM_VERSION} as ci-package-pyside
+# We are currently not interested in the Python implementation and bindings
```
Simran-B, line 12: Not sure if Python is really not needed. Need to investigate whether something runs the Python interpreter or if there's always some Python supported built into the library.

Simran-B: Adapters are all Python-based, and there is no real plan to port them to C++. We don't use adapters, however.

cmake/FindOpenColorIO.cmake
Simran-B: New dep package has the DLL under its original name.

PR #1492, opened 2021-02-20
Add new translations for German localisation (WIP)

mara004:
I've worked on extending the German localisation for Olive. Although it is not finished or tested yet, I wanted to share what I translated so far to prevent the same work from being done twice by different people ;)
By the way: The compile and translate instructions are great!
Kind regards & thank you for developing Olive


mara004:
I would probably need some help:
I ran `lupdate` as recommended in the wiki, which I already did before and worked well. However, this time, it seems like `lupdate` "destroyed" the code path / declaration references... The translation still compiles and loads correctly, but that is not nice. How can I revert this? Diff with a previous version, maybe?
For testing purposes, I ran lupdate on two other languages, and it also removed the declarations.

mara004:
Okay, I did the diff/merge and restored what got lost. However, I can't help but find `lupdate` questionable now...

Simran-B:
So in 412864d2aa8283986af2d8f0f26bc55c23d8bd26 it removed all the `<location>` elements. That is peculiar indeed. The only way I get lupdate to strip the code locations is with `lupdate . -locations none -ts app/ts/de_DE.ts`... Was there any difference in the way you invoked lupdate, like from a different working directory or shell? Do you have multiple Qt versions installed by any chance?

This change in the parent commit 9478cb17e1f25e098758c45d87c983f8bd485ac1 is also really strange:

image

Single quotes are invalid for wrapping attribute values in XML. Did lupdate do this?

> things destroyed by lupdate and KDE Lokalize's special character re-encoding

What happened to special characters? Qt should have no trouble with UTF-8. Can you double-check whether lupdate is at fault here? I haven't seen any special character issues with lupdate + linguist under Windows.

mara004:
Yeah, it was really strange. It's possible that I inadvertently called `lupdate` from the wrong directory.
No, I don't have multiple Qt versions installed but I'm on KDE Neon which is based on Ubuntu Focal and overrides KDE software with newer versions. This kind of repository management can sometimes have issues.
I reverted that change in the parent commit, I think it happened because of Lokalize (https://apps.kde.org/en/lokalize), which apparently has numerous strange issues regarding .ts files.
The changes of some characters also came from Lokalize (it replaced all ' characters with &apos; " with &quot; and & with &amp;), but it was not an issue with non-UTF8 characters like umlauts (I was unclear in the commit description, sorry). Those character changes probably didn't even harm, but I reverted them as they made the plain xml file less well readable.
Thanks for your detailed investigations!

mara004:
Yeah, it was really strange. It's possible that I inadvertently called `lupdate` from the wrong directory.
No, I don't have multiple Qt versions installed but I'm on KDE Neon which is based on Ubuntu Focal and overrides KDE software with newer versions. This kind of repository management can sometimes have issues.
I reverted that change in the parent commit, I think it happened because of Lokalize (`https://apps.kde.org/en/lokalize`), which apparently has numerous strange issues regarding .ts files.
The changes of some characters also came from Lokalize (it replaced all `'` characters with `&apos;`, `"` with `&quot;`, and `&` with `&amp;`), but it was not an issue with non-UTF8 characters like umlauts (I was unclear in the commit description, sorry). What I found suspicious about those character replacings is that it did not only change translations, but also source texts, which is why I reverted it.
Thanks for your detailed investigation and sorry for all this confusion!

Simran-B:
I see, these substitutions are an encoding scheme that you also find in HTML (keyword HTML entities). I would have to look up the XML spec to check when it is strictly necessary to encode reserved characters. lupdate and linguist seem to be fine with them unencoded though.

mara004:
I have some trouble again: I rebuilt from a57bf35e, and now the translations for the `Crop` node don't appear, even though everything related to crop is translated in Linguist. I also ran lupdate on the translation after pulling from git (which works well now as I called it from the right directory).

Simran-B:
Can you modify `app/node/distort/crop/cropdistortnode.h` like shown below and see if that fixes it?

```diff
@@ -30,6 +30,7 @@ namespace olive {
 
 class CropDistortNode : public Node
 {
+  Q_OBJECT
 public:
   CropDistortNode();
```

mara004:
Indeed. That fixed it.

mara004:
Again some little translation issue:
If I open up Olive and didn't do an action yet, then the strings Undo / Redo aren't translated. However, as soon as I do an undoable action, both translations show up properly.
Screenshots:
Screenshot_20210309_112211
Screenshot_20210309_112144

itsmattkc:
I wonder if it would be better to forego `createUndoAction` and `createRedoAction` and just make our own custom actions to keep it streamlined/consistent?

Simran-B:
Whatever works. I don't understand Qt's undo framework well enough to judge. If there's a way to control/update the translations of Undo/Redo then that would also suffice.

mara004:
I wonder whether it's possible to skip the CI? It seems somewhat wasteful to me to always trigger a build when I just update the translation...

Simran-B:
It would be possible to exclude the translation folder, but you can also put `[skip ci]` in the commit message whenever you don't want an automated build.

https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

mara004:
@Simran-B Could you please review the translation? I might have made similar mistakes as the Italian translator.
In case this pull request gets merged some day, should we maybe rebase it somehow in order not to blow up the git history?

Simran-B:
Looks fine, the changes to the context seem to be caused by re-ordering of translation strings or because some went away or got renamed, but that's expected. I'm going to merge the translations, they look solid. Thanks! There are some more things that could use tweaking, but we can always improve things later.

mara004:
Thanks!

PR #1534, opened 2021-03-24
Update cs_CZ.ts

pafri:
Completion of the file - updated via lupdate command.


Simran-B:
Resolved conflicts with #1398 as good as possible, merging.

PR #1552, opened 2021-04-06
Add extra mouse cursors i.e. hand cursors (and tool suggestion)

VSchris-code:
This is more detailed than the previous pull, which was denied.
So, here's the point.
The open and closed hands can be used to mark something draggable in a effect's settings panel and when you're dragging a setting that can be dragged or a clip in the timeline respectively- and they should appear in Windows only.
The pointing hand cursor appears when your cursor is on a clip's FX button or crop button, much like how Sony Vegas Pro does, much like the open and closed hands- they can only appear on Windows.
A variant of the pointing hand cursor with arrows pointing to the left and right to the left and right of the index finger appears when you're changing a number setting with the move left and right method, much like in Adobe Photoshop and Animate (and various other software by Adobe)- appearing on all three supported platforms.
And also a suggestion for a pre-existing cursor: Make the razor cursor look a bit more like a actual razor cutter.
And also a delete tool, which, deletes stuff from the timeline. Nothing more.
The delete tool's own cursor would be a tailless/tailed arrow with an X next to it- and the delete tool can be more useful than regular Delete- or Backspace-key deletion because when you hold down LMB with the delete tool on an empty part of the timeline, you can select multiple things to delete from the timeline, or you can just hold either Control key, and you can click on whatever things you want to delete from the timeline, and then release Control, and then poof. They're gone.

Whew, quite a lot of stuff to type for a tool suggestion. Now for more cursor suggestions...
A rotation cursor for when you're rotating something. It's that simple.


elsandosgrande:
@VSchris-code 

1. From everything that I can see here, you're trying to pull from the master branch into the 0.1.x branch. That's never going to happen.
2. Windows-only features are almost certainly never going to happen (there may possibly be exceptions in the future, but the UI is never going to be one of them).
3. 0.1.x is **deprecated**, full stop. **Nothing new is going to be added to it.**
4. Things like pointer icons have the lowest priority right now, since getting the node editor and cache into shape are top priorities, as almost everything else hinges on them.

Simran-B:
@VSchris-code Stop opening pointless pull requests or you will be blocked permanently. Pull requests are for actual code contributions, not feature requests. If you have something to suggest then create an issue, but be warned that we don't accepting feature requests at the moment #1175

PR #1562, opened 2021-04-12
Fixes for OTIO merges

ThomasWilshaw:
I hadn't fully tested the recetly merged OTIO code with the latest master. This branch aims to sort out/fix those changes.

### DIFF
--- a/app/core.cpp
+++ b/app/core.cpp
@@ -1501,14 +1501,6 @@ void Core::CacheActiveSequence(bool in_out_only)
 bool Core::ValidateFootageInLoadedProject(Project* project, const QString& project_saved_url)
 {
   QVector<Footage*> project_footage = project->root()->ListChildrenOfType<Footage>();
-
-  // OTIO files import their footage into a folder so check them too
-  foreach(Folder * folder, project->root()->ListChildrenOfType<Folder>()) {
-    QVector<Footage*> footage = folder->ListChildrenOfType<Footage>();
-    if (!footage.isEmpty()) {
-      project_footage.append(footage);
-    }
-  }
   QVector<Footage*> footage_we_couldnt_validate;
 
   foreach (Footage* footage, project_footage) {
--- a/app/core.cpp
+++ b/app/core.cpp
@@ -1501,6 +1501,14 @@ void Core::CacheActiveSequence(bool in_out_only)
 bool Core::ValidateFootageInLoadedProject(Project* project, const QString& project_saved_url)
 {
   QVector<Footage*> project_footage = project->root()->ListChildrenOfType<Footage>();
+
+  // OTIO files import their footage into a folder so check them too
+  foreach(Folder * folder, project->root()->ListChildrenOfType<Folder>()) {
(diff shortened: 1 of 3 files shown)


app/node/project/footage/footage.cpp
```diff
@@ -36,6 +36,7 @@
 namespace olive {
 
 const QString Footage::kFilenameInput = QStringLiteral("file_in");
+const QString Footage::kVideoParamsInput = QStringLiteral("video_params");
```
itsmattkc, line 39: Hmm? Footage is (now) a derivative of ViewerOutput which defines `kVideoParamsInput` here: https://github.com/olive-editor/olive/blob/master/app/node/output/viewer/viewer.cpp#L30 (it would fail to compile if it wasn't defined)

Seems more likely the crash is elsewhere and this "dodges" it by making the connection somewhere else.

ThomasWilshaw: So it is, I missed that.

As far as I can tell if I load an OTIO file with missing media Olive doesn't created a VideoParams widget. When I then relink the footage it tries to refernece something that doesn't exist. I'm not reall sure how to fix this.

ThomasWilshaw:
I think these changes are ready for review. There's still work to be done on the OTIO stuff but I think these need to be added before I do that.

elsandosgrande:
Is this request still active? I mean, the review was requested on the sixth of May and there has been no activity since.

itsmattkc:
Ahh I remember this, I was going to make `ListChildrenOfType` recursive before merging this so that this PR doesn't have to check folders separately (it should be recursive anyway because otherwise folders in users' projects won't work either).

I have just done this in 5254be3fb5d3755745a0b2612331c1c29300cc0f, so if the rest of this PR is still good, I think it's ready to merge.

ThomasWilshaw:
If I recall correctly there was also the issue whereby if I set autoposition to true I'd get crashes, I donlt know if any of that was changed in your recent node viewer branch

ThomasWilshaw:
@itsmattkc I think this is ready to be added now if it all looks okay to you. I think it puts OTIO at a very basic usable state. There's still stuff to do (edge cases, the other OITO PR and testing) but it's probably best if this is in first. If you're happy with everything I'll squash it before you add it as it's quite messy.

itsmattkc:
For the most part, this looks pretty good, however using MultiUndoCommands is unnecessary since this is loading a project and will theoretically allow much of this to be undone by the user which feels undesirable.

Most of the functions have "non-undoable" variants you can call, for example adding a node to a graph is done automatically when you set the node's parent to the graph (`node->setParent(graph);`), you can set positions with `NodeGraph::SetNodePosition`, and you can connect nodes with `Node::ConnectEdge`. Alternatively, you can keep them as-is and just call `redo_now()` on the commands instead of pushing them to the stack (like the FolderAddCommands).

ThomasWilshaw:
Thanks, I've done that.

itsmattkc:
Sorry for the delay, seems all good

PR #1575, opened 2021-04-14
nodeviewitem: Add flag for drawing title arrow

ThomasWilshaw:
The collapse arrow was being drawn twice, add a flag that defaults to true to select whether DrawNodeTitle draws said arrow.

### DIFF
--- a/app/widget/nodeview/nodeviewitem.cpp
+++ b/app/widget/nodeview/nodeviewitem.cpp
@@ -296,7 +296,7 @@ void NodeViewItem::paint(QPainter *painter, const QStyleOptionGraphicsItem *opti
       DrawNodeTitle(painter, node_label, safe_label_bounds, Qt::AlignTop);
       f.setPointSizeF(font_sz * 0.6);
       painter->setFont(f);
-      DrawNodeTitle(painter, node_shortname, safe_label_bounds, Qt::AlignBottom);
+      DrawNodeTitle(painter, node_shortname, safe_label_bounds, Qt::AlignBottom, false);
     }
 
   }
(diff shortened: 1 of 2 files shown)


app/widget/nodeview/nodeviewitem.cpp
```diff
@@ -367,22 +367,24 @@ void NodeViewItem::ReadjustAllEdges()
...
-  int icon_size = fm.height() / 2;
-  int icon_padding = title_bar_rect_.height() / 2 - icon_size / 2;
-  int icon_full_size = icon_size + icon_padding * 2;
-  const QIcon& expand_icon = IsExpanded() ? icon::TriDown : icon::TriRight;
-  int icon_size_scaled = icon_size * painter->transform().m11();
-  painter->drawPixmap(QRect(title_bar_rect_.x() + icon_padding,
-                            title_bar_rect_.y() + icon_padding,
-                            icon_size,
-                            icon_size), expand_icon.pixmap(QSize(icon_size_scaled, icon_size_scaled)));
+  int icon_full_size = 0;
```
itsmattkc, line 377: Hmm wouldn't setting `icon_full_size` to 0 mess up the calculations that it's used for later on?

itsmattkc (changes requested):

app/widget/nodeview/nodeviewitem.cpp
ThomasWilshaw: I see what you mean, I forgot there'll always be an arrow if if this function hasn't drawn it. Moving the relevant calculations outside the if statement should do it right? Or will other arrow being a bit bigger cause problems?

itsmattkc: Ahh that's a good point, the size difference definitely makes it challenging. It might actually be fine as-is since the small subtitle should never be too wide for the node (though I guess we can't guarantee that through every language). Not sure what the best solution is in that case, outside of maybe passing along the arrow size to the function independently of the current font size.

PR #1587, opened 2021-04-24
Small audio caching optimisation

ThomasWilshaw:
A couple of changes based of profiling while caching. An example of the improvement form the first commit:
image

The second commit gave a similar improvement.

### DIFF
--- a/app/render/audioparams.h
+++ b/app/render/audioparams.h
@@ -66,6 +66,9 @@ public:
     format_(kFormatInvalid)
   {
     set_default_footage_parameters();
+
+    // Cache channel count
+    channel_count_ = av_get_channel_layout_nb_channels(channel_layout());
   }
 
   AudioParams(const int& sample_rate, const uint64_t& channel_layout, const Format& format) :
@@ -75,6 +78,9 @@ public:
   {
     set_default_footage_parameters();
     timebase_ = sample_rate_as_time_base();
+
+    // Cache channel count
+    channel_count_ = av_get_channel_layout_nb_channels(this->channel_layout());
   }
 
   int sample_rate() const
@@ -200,6 +206,8 @@ private:
   int sample_rate_;
 
   uint64_t channel_layout_;
+
+  int channel_count_;
 
   Format format_;
 
--- a/app/render/audioparams.h
+++ b/app/render/audioparams.h
@@ -68,7 +68,7 @@ public:
     set_default_footage_parameters();
 
     // Cache channel count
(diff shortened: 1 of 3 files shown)


app/render/audioparams.h
```diff
@@ -75,6 +78,9 @@ class AudioParams {
   {
     set_default_footage_parameters();
     timebase_ = sample_rate_as_time_base();
+
+    // Cache channel count
+    channel_count_ = av_get_channel_layout_nb_channels(this->channel_layout());
```
itsmattkc, line 83: You'll also want to set this under `set_channel_layout()`, and it'd probably be best to put this under a new function like `calculate_channel_count` to minimize on copy/pasted code (see `VideoParams:: calculate_effective_size`).

app/render/audioplaybackcache.cpp
```diff
@@ -64,10 +64,14 @@ void AudioPlaybackCache::WritePCM(const TimeRange &range, SampleBufferPtr sample
...
-    playlist_.push_back(CreateSegment(seg_sz, playlist_.GetLength()));
-    length_diff -= seg_sz;
+  qint64 whole_segments = length_diff / kDefaultSegmentSize;
+  qint64 remainder = length_diff % kDefaultSegmentSize;
+
+  for (int i = 0; i < whole_segments; i++) {
+    playlist_.push_back(CreateSegment(kDefaultSegmentSize, playlist_.GetLength()));
+  }
+  if (remainder > 0) {
+    playlist_.push_back(CreateSegment(remainder, playlist_.GetLength()));
```
itsmattkc, line 74: Does this actually make a difference? It looks like it does the exact same thing but with more lines of code.

itsmattkc (changes requested):

app/render/audioparams.h
ThomasWilshaw: Will do, thanks.

app/render/audioplaybackcache.cpp
ThomasWilshaw: It did seem too, I'll try and redo the tests and post them here. As I understand it, this removes the repeated qmin and probably allows a bit of branch prediction with the if statement. Happy to be corrected though :)

itsmattkc: Hmm perhaps. Yes I'll be interested to see the test information.

ThomasWilshaw: I can't actually get a repeat test so I suspect it was a fluke, shall I revert the commit?

PR #1588, opened 2021-04-24
Update standard test name to improve usability in Visual Studio.

ThomasWilshaw:
This change just improves the GUI experience of running CTests in Visual Studio. I think it interprets the name as `[namespace].[class].[name]`. There's a known bug in VS (https://developercommunity.visualstudio.com/t/cmake-weird-test-name-prefix/793209) where it appends a random generated string to the start of all tests which makes searching through them a pain and this should make it easier to filter tests. It's not a big issue with the current three tests but could be going forwards.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -62,7 +62,7 @@ function(olive_add_test GROUP NAME SOURCE)
     PRIVATE
     ${OLIVE_COMPILE_OPTIONS}
   )
-  add_test(${NAME} ${NAME})
+  add_test("OliveLib.Tests.${NAME}" ${NAME})
 endfunction()
 
 add_subdirectory(compositing)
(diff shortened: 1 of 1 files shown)


itsmattkc:
How about this instead:
```
if (MSVC)
  add_test("Olive.${GROUP}.${NAME}" ${NAME})
else()
  add_test(${NAME} ${NAME})
endif()
```

ThomasWilshaw:
Thanks, CMake is still a wormhole for me. What does `${GROUP}` return? Do you think there's an issue with having inconsistent tests between platforms?

itsmattkc:
> What does ${GROUP} return?

`${GROUP}` is just one of the function's arguments, currently other files use "Timeline", "General", and "Compositing" as group names. It should just be a little more descriptive.

> Do you think there's an issue with having inconsistent tests between platforms?

Mostly my feeling is that if this is an issue that predominantly affects MSVC, than the fix should target MSVC too.

ThomasWilshaw:
Thanks, that works a lot better.

PR #1589, opened 2021-04-26
Update zh_CN.ts

XGSZZ:


itsmattkc:
Does this PR depend on #1578 or is it completely independent?

XGSZZ:
> 
> 
> Does this PR depend on #1578 or is it completely independent?

It's independent.

PR #1674, opened 2021-07-17
Japanese translation

yoshiomiyamae:
I added Japanese translation. Please pull it.
Basically almost all translation has finished, but I am not familiar with Olive yet, so I may change something later.


itsmattkc:
Thanks for the translation!

PR #1676, opened 2021-07-19
Add H.264 RGB export option

CobaltSpace:
x264 does support RGB colorspace, but ffmpeg has RGB separated as `libx264rgb`.  There have been mentions of including codecs such as `ffv1`, but in my experience, `x264` with `qp=0` or `crf=0` is faster and compresses better than `ffv1`.

The changed files were run through `clang-format`.


app/dialog/export/exportvideotab.cpp
```diff
@@ -49,28 +46,23 @@ ExportVideoTab::ExportVideoTab(ColorManager* color_manager, QWidget *parent) :
...
+int ExportVideoTab::SetFormat(ExportFormat::Format format) {
   format_ = format;
 
   QList<ExportCodec::Codec> vcodecs = ExportFormat::GetVideoCodecs(format);
   setEnabled(!vcodecs.isEmpty());
   codec_combobox()->clear();
-  foreach (ExportCodec::Codec vcodec, vcodecs) {
-    codec_combobox()->addItem(ExportCodec::GetCodecName(vcodec), vcodec);
-  }
+  foreach (ExportCodec::Codec vcodec, vcodecs) { codec_combobox()->addItem(ExportCodec::GetCodecName(vcodec), vcodec); }
```
Simran-B, line 55: That's a really questionable formatting choice 🤔

Simran-B:
I saw the following in another repo:

`libx264rgb_encoder_deps="libx264 x264_csp_bgr"`

Does this PR require additional dependencies or specific versions of them, or does x264 come with these by default?

CobaltSpace:
I don't think any new dependencies are required.

app/dialog/export/exportvideotab.cpp
CobaltSpace: I guess maybe I shouldn't have just let clang-format run free, considering after checking the `.clang-format` file, I realize that it has set the column limit to 120, while `CONTRIBUTING.md` says 100.  I now tried reformatting with it set to 100, and I can see why that is probably not a good idea, so I will go through and fix the ones that are problematic if that is preferred.

Simran-B: We are actually not sure how that `.clang-format` was created and how it got added, but it doesn't match the actual code style of 0.2, which follows Google's conventions for the most part. We will try to adjust the config to match the current code style.

CobaltSpace:
This pull request should not rely on clang-format.  I removed all the formatting changes so only the code changes remain.  I made my changes match the formatting that was already in use surrounding them.

PR #1677, opened 2021-07-19
Update code style formatting configuration (clang-format)

Simran-B:
Based on Google style as of clang-format v10.0.1

`ColumnLimit` raised from 80 to 100 (unclear where 120 came from)

CMake integration inspired by https://arcanis.me/en/2015/10/17/cppcheck-and-clang-format#clang-cmake

Formatting the code with `cmake --build . --target format` appears to affect all files. It looks like the current code style is different from Google's in various smaller aspects.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,16 +3,20 @@ Language:        Cpp
 # BasedOnStyle:  Google
 AccessModifierOffset: -1
 AlignAfterOpenBracket: Align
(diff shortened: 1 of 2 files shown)


ThomasWilshaw:

Simran-B:
Still relevant and the refactoring doesn't have much of an impact on this work, except that the style should be enforced in all repos.

Issue #1345, opened 2020-11-18
[TODO] Finish implementing export formats that aren't H.264


itsmattkc:
_Originally opened as https://github.com/sobotka/olive/issues/76_

**General TODO Summary**
Ensure the output codecs that we do support actually work properly.

**Additional Information / Notes**
The export dialog offers codecs such as H.265, ProRes, DNxHD, image sequences, etc. however only H.264 has actually been tested. The rest of the codecs have extremely varying degrees of usefulness and should at some point be tested to output as expected.

itsmattkc:
<a href="https://github.com/sobotka"><img src="https://avatars1.githubusercontent.com/u/59577?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [sobotka](https://github.com/sobotka)**
_Saturday May 02, 2020 at 15:15 GMT_

----

Only note here is to handle each codec on a case by case basis to properly and robustly support them, within the design constraints of an NLE.

General Do:
* Do “generally” support encoding characteristics. For example, ProRes offers the standardized and general categories of output. Support those.
* Do keep the encoder to useful, but perhaps not “industrial”. This is a fuzzy line of qualia of course, but *O*live *I*s *N*ot *A*n *E*ncoder. As such, trying to keep a “sane” hand on encoding quality, while at the same time acknowledging that Olive can export native, raw frames, of the highest possible quality, provides the audience with a means to take Olive footage elsewhere for encoding.
* Do support limited codecs, and support them sanely and well. The more codecs that are supported, the more tech overhead and nightmare unfolds.

General Don’ts:
* Don’t support general things across all codecs and then arbitrarily not have feedback in the UI where they are or are not exposed due to the nature of the codec. Plenty of software does this, and it is nightmare.
* Don’t try to support more codecs than the most baseline offerings suitable for editing. Robust support of DNxHD, ProRes, H264, H265, and still image EXR, DPX, TIFF cover a majority of bases. Some require additional colourimetry handling etc. As such, handling fewer robustly versus many half-assedly.

A decent example of “robust” handling is Fusion’s handling of DPX. It is a nuanced still image encoding format, but it shows the complexity of handling it with the details required.

itsmattkc:
<a href="https://github.com/itsmattkc"><img src="https://avatars1.githubusercontent.com/u/34096995?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [itsmattkc](https://github.com/itsmattkc)**
[...]

PR #1678, opened 2021-07-19
Add webm support, useful for incrustation of transparent videos

ghpu:
Webm/vp9 is one of the rare video codecs supporting transparent background.
Interesting for making animated logos.


Simran-B:
Thanks for sharing this. 0.1.x is unfortunately not maintained anymore.

QuickTime Animation (RLE) supports an alpha channel and is over two decades old. The "QuickTime" RGBA option in AfterEffects CS6 uses this format for instance. There are also Apple ProRes, Cineon, DPX, etc. WebM/VPX has basically no relevance outside of the browser.

Also see the discussion in #1345

PR #1680, opened 2021-07-20
Add .clang-tidy for better style and best practices linting

CobaltSpace:
> **clang-tidy** is a clang-based C++ “linter” tool. Its purpose is to provide an extensible framework for diagnosing and fixing typical programming errors, like style violations, interface misuse, or bugs that can be deduced via static analysis. **clang-tidy** is modular and provides a convenient interface for writing new checks.

With a `.clang-tidy`, text editors and IDE's that have integration with clang-tidy (or clangd) will be able to complain about anti-patterns, and it can also help with enforcing code style in ways that clang-format can't.  clang-tidy is basically a better clang-check.

The file currently has what I think is probably a good starting point, which is turning all checks on and then turning off some of the checks that conflict with what has been done.  The choice of checks to use is pretty subjective, so project leads may want to refer to https://clang.llvm.org/extra/clang-tidy/checks/list.html and choose which ones they will want to disable/enable.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,12 @@
+---
+# Use all checks except for conflicts with project style.
+Checks: "*,\
+-darwin-*,\
+-fuchsia-*,\
+-llvmlibc-*,\
+-modernize-use-trailing-return-type,\
+fuchsia-trailing-return,\
+-llvm-qualified-auto,\
+-readability-qualified-auto,\
+-readability-implicit-bool-conversion"
+...


elsandosgrande:
@Simran-B What's your comment on this with regards to your work in #1677?

Simran-B:
@elsandosgrande It's not related.

elsandosgrande:
> and it can also help with enforcing code style in ways that clang-format can't.

This is what piqued my interest @Simran-B.

Eh, I'll look into `clang-tidy` and `clang-format` later and compare the two in VS Code with the appropriate extensions.

CobaltSpace:
`clang-tidy` & `clang-format` should not be compared as if they address the same issues.  They do not.  They can both be used.  I would argue focusing on #1677 is more important than this.

`clang-tidy` does have "fixes" for some of the checks it does, but if automatic fixing is to be used, then I believe that maintainers and leads will want to really want to configure the `.clang-tidy` file to their liking.  I recommend against using `clang-tidy`'s automatic fixing until then, and just let `clang-tidy` be there to say things like "that can cause problems, consider using this instead".

elsandosgrande:
Mhm. So, as far as I can tell from the LLVM documentation — it really needs a dedicated page discussing the differences between the two in detail — `clang-tidy` is akin to a C++ linter tool that doesn't really deal with formatting, whereas `clang-format` only deals with formatting and nothing more. Please correct me if I'm wrong by the way, as there doesn't seem to be any direct comparison between the two online that clears up the confusion and the Clang documentation doesn't give examples of what each tool does.

PR #1711, opened 2021-08-09
Fix #1708 Node edge arrow in wrong place

### LINKED ISSUE #1708: [NODES] Small triangle appears when Ctrl clicking a node
**Commit Hash** 
132e637f

**Platform** 
Windows 10

**Summary**
When I Ctrl click a node to create a new edge a small grey triangle appears in the node view.

At first it only happens in the track view context, but once I've been in there once or selected "view all nodes" it happens in any context.
The triangle always appears in the same place ina a given context, whatever node I click. THe triangle disapears as soon as I move the mouse.

image

ThomasWilshaw:
Fixes #1708 

When a node edge is first created its from_item and to_item are not
initilised. This was leading to the arrow at the end of the edge being
drawn in a random place. The work around here is to explicitly set the
arrow to be drawn only when the mouse has started moving and we know
where to draw it.

### DIFF
--- a/app/widget/nodeview/nodeview.cpp
+++ b/app/widget/nodeview/nodeview.cpp
@@ -563,6 +563,7 @@ void NodeView::mouseMoveEvent(QMouseEvent *event)
       create_edge_->SetPoints(create_edge_src_->GetOutputPoint(Node::kDefaultOutput),
                               scene_pt,
                               false);
+      create_edge_->SetDrawArrow(true);
     }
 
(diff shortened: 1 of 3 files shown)


itsmattkc:
The cause of this was the edge not being positioned until the mouse moved, which would leave it at (0, 0) on mouse press. I decided a more robust solution would be to just position it immediately on creation.

ThomasWilshaw:
That makes a lot more sense :)

Issue #1690, opened 2021-07-27
[EDIT] Timeline and Footage Markers


unfa:
**Commit Hash** f13962da

**Platform** Manjaro Linux KDE

**Summary**

Various issues related to Markers

- [x] It's not possible to select markers
- [x] Hence it's not possible to delete
- [x] Move markers
- [x] It's also not possible to rename markers
- [x] Marker names are not visible on the timeline ruler or anywhere else
- [x] Markers created in Media preview do not show up on clips in a sequence
- [x] There's no way to change marker color (always green)

itsmattkc:
Some of this overlaps with #487, but this post is probably more complete. May merge that issue into this one.

itsmattkc:
This has now been fully implemented

PR #1717, opened 2021-08-11
Make markers in to widgets in their own right

ThomasWilshaw:
This is a first attempt to turn markers into widgets (#1690). At this point I'm mainly looking ot see if I'm going in the right direction or not.

I've created a marker widget that can change color and started to hook it up to `seekablewidget`. It's still very scrappy but sort of works already.

For my own references:
-[Draw markers](https://github.com/olive-editor/olive/blob/9f4470f55cb799313b46cc0802535d627f8ac621/app/widget/timeruler/seekablewidget.cpp#L165)
-[Timeruler specific code](https://github.com/olive-editor/olive/blob/2f805abad8f59e58074f9284f69d8bc525a8359b/app/widget/timeruler/timeruler.cpp#L107l)


itsmattkc:
Mostly this looks to be on the right track, I wouldn't have thought of using QWidgets but it's not a bad approach at all. I'll post some comments on the parts I think should be revised.

app/widget/marker/marker.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+  int y = text_height; //13
+
+  int half_width = marker_width_ / 2;
+
+  int x = half_width; //3
+
+  if (x + half_width < 0 || x - half_width > width()) {
+    return;
+  }
```
itsmattkc, line 61: Don't need to worry about filtering code like this, Qt will automatically skip painting markers that are offscreen

app/widget/marker/marker.cpp
```diff
@@ -0,0 +1,100 @@
...
+	QWidget(parent),
+    marker_color_(Qt::green)
+{
+  setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed);
+  setMinimumSize(20, 20);
+  setContextMenuPolicy(Qt::CustomContextMenu);
+
+  connect(this, &Marker::customContextMenuRequested, this, &Marker::ShowContextMenu);
+
+  color_coding_menu_ = new ColorLabelMenu();
```
itsmattkc, line 42: Not giving this menu a parent will create a memory leak. I'd be more inclined to make this object on the stack in `ShowContextMenu` and connect it there too just out of ease.

app/widget/marker/marker.cpp
```diff
@@ -0,0 +1,100 @@
...
+      QPoint(x - half_width, y - half_text_height),
+      QPoint(x - half_width, y - text_height),
+      QPoint(x + 1 + half_width, y - text_height),
+      QPoint(x + 1 + half_width, y - half_text_height),
+      QPoint(x + 1, y),
+  };
+
+  p.drawPolygon(points, 6);
+
+  p.setRenderHint(QPainter::Antialiasing, false);
```
itsmattkc, line 82: Disabling the antialiasing isn't necessary since the QPainter gets destroyed immediately after this function ends.

app/widget/marker/marker.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "common/qtutils.h"
+#include "ui/colorcoding.h"
+#include "widget/menu/menu.h"
+#include "widget/menu/menushared.h"
+
+namespace olive {
+
+Marker::Marker(QWidget *parent) :
+	QWidget(parent),
+    marker_color_(Qt::green)
```
itsmattkc, line 34: Would probably make `marker_color_` an integer and start this with one of the predetermined colors. That way it'll all stay consistent and serialize into the project files better.

app/widget/timeruler/seekablewidget.cpp
```diff
@@ -127,6 +133,23 @@ void SeekableWidget::SetScroll(int s)
...
+    marker_widgets.clear();
+   
+    foreach(TimelineMarker * marker, timeline_points()->markers()->list()) {
+      Marker* marker_widget = new Marker();
+      marker_widgets.append(marker_widget);
+      marker_layout_->addWidget(marker_widget);
+    }
+  }
+}
+
```
itsmattkc, line 152: Would probably be inclined to have separate `MarkerAdd` and `MarkerRemove` functions that just work with one marker at a time, rather than destroying and creating them all each time. I often use a QMap or QHash to match source objects with their widgets (see NodeViewScene::item_map_)

app/widget/timeruler/seekablewidget.cpp
```diff
@@ -44,6 +44,10 @@ SeekableWidget::SeekableWidget(QWidget* parent) :
   playhead_width_ = QtUtils::QFontMetricsWidth(fm, "H");
 
   setContextMenuPolicy(Qt::CustomContextMenu);
+
+  marker_layout_ = new QHBoxLayout(this);
```
itsmattkc, line 48: Wouldn't bother with the QHBoxLayout, I think you'll most likely have to manually control their positions anyway

itsmattkc (commented):
A lot of these things you were probably already planning to do, so apologies for any redundancy in that regard. Just wanted to get all my thoughts across

ThomasWilshaw:
Thanks for the review, they're all very useful points. I'll respond to each point as I work through them.  Out of curiosty what would you have used instead of a QWidget?

app/widget/timeruler/seekablewidget.cpp
ThomasWilshaw: I've added a basic add marker function. How owuld you want to delete a marker? I suppose there needs to be some UI indication that it's been selected. Would you use the delete key or right click->delete?

app/widget/marker/marker.cpp
ThomasWilshaw: Could you check I've done this right?

app/widget/timeruler/seekablewidget.cpp
itsmattkc: I think you should add the "edit menu items" (from MenuShared) to the context menu, which will add things like undo, redo, delete, cut, copy, paste, etc. Then add some logic for TimeBasedWidget to pass through from functions like `DeleteSelected` to the TimeRuler. Functions like "HandPress" and "PlayheadPress" might provide a good reference for implementing that. Doing that will ensure both right click > delete and the delete key will work.

app/widget/marker/marker.cpp
itsmattkc: The new `ShowContextMenu`? Seems good to me

itsmattkc:
> Out of curiosty what would you have used instead of a QWidget?

I had considered switching to a QGraphicsView, but it's probably overkill for this.

app/widget/timeruler/seekablewidget.cpp
ThomasWilshaw: So I'm going to have to implement a bunch of AddMarkerUndo Command type classes then as well?

If I copy a marker, I assume all I'm copying is the colour and label? You wouldn't want the time/position data would you?

app/widget/timebased/timebasedwidget.h
```diff
@@ -54,6 +54,28 @@ class TimeBasedWidget : public TimelineScaledWidget
 
   virtual bool eventFilter(QObject* object, QEvent* event) override;
 
+  // Temp file location
```
ThomasWilshaw, line 57: @itsmattkc Do you think it would make sense to move this and the marker add class to their own `markerundo.h` file alongside `marker.h`?

ThomasWilshaw:
To do:

- [ ] Add warning box for malformed paste data
- [ ] Figure out duplicate and insert
- [ ] Make markers renameable and show names
- [ ] Make markers moveable?
- [ ] Select on left click
- [ ] Time range markers (in/out points)
- [ ] ...

PR #1722, opened 2021-08-16
Tweak pure red/green cache colors to better fusion with styles

pgilfernandez:
Tweak pure red/green cache colors to better look with any style.

Old vs New:
cache_color

### DIFF
--- a/app/widget/timeruler/timeruler.cpp
+++ b/app/widget/timeruler/timeruler.cpp
@@ -262,7 +262,7 @@ void TimeRuler::paintEvent(QPaintEvent *)
     if (cache_screen_length > 0) {
       int cache_y = height() - cache_status_height_;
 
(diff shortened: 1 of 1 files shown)


itsmattkc:
I'm not sure I'd want to change these (this and the cache indicator). These have very specific functions for the user and I think the high contrast helps with their function.

pgilfernandez:
IMHO pure colors could "hurt" the eyes, hehehe (just kidding), while the colors I'm proposing are still red and green, they are softer but still significant enough to be functional... give them a try, they work very well while you use Olive.

itsmattkc:
I'm not sure I like how muted it is, and using other "profesionally-targeted" applications as reference, while they might not be using "pure" red/yellow/green, their audio monitors are all pretty bright. Once again, perhaps I'd be open to something more in-between.

mixers

(Premiere Pro, Pro Tools, and Logic Pro X respectively)

PR #1724, opened 2021-08-16
Tweak Color coding

pgilfernandez:
Tweak bright "color table" to a more relaxed ones to use in Nodes and Clips.

Old vs New:
colors

Maybe some color names need to be updated as well, but I'm unsure about it...


itsmattkc:
Ahh good, I was hoping someone would do this. The existing colors are certainly crude and I never intended for them to be permanent.

itsmattkc:
Actually, would you be able to tweak them to be a little more saturated? While I agree the previous colors were crude, these have a tendency to look a little samey when they should provide important differentiation to the user:

image

Something in between what was before and these ones would be ideal IMHO.

pgilfernandez:
Indeed, I will do it right now

PR #1759, opened 2021-10-02
Testing vcpkg action

itsmattkc:
Just testing the viability of this.


ThomasWilshaw:
If you're running into space issues this might help:
https://github.com/Microsoft/vcpkg/issues/2874#issuecomment-521353560

But is suspect the ~2 hour build time is more of an issue.

itsmattkc:
> If you're running into space issues this might help:

Ah, thanks for the tip, I'll try that.

> But is suspect the ~2 hour build time is more of an issue.

The GitHub Action we use is supposed to cache the results, so presumably after the first run, it shouldn't take long at all.

PR #1770, opened 2021-10-13
AudioVisualWaveform, SumSamples: improve performance by 10x on x86

morrolinux:
Improve performance by 10x on x86 architecture by leveraging platform specific optimizations.
This will show in real-world application when we'll switch to data parallel paradigm for jobs

### DIFF
--- a/app/node/math/math/mathbase.cpp
+++ b/app/node/math/math/mathbase.cpp
@@ -175,7 +175,6 @@ void MathNodeBase::PerformAllOnFloatBuffer(Operation operation, float *a, float 
 }
 
 #ifdef Q_PROCESSOR_X86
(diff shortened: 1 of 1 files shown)


Simran-B:
What machine code do compilers emit without these optimizations? GCC tends to be very aggressive in auto-vectorization.

Simran-B:
It's definitely a huge step up in performance and I agree that using SSE consistently is better than to rely on compilers. Speaking of which, using O2 for development makes sense IMO but we could use O3 for release builds, couldn't we? A benchmark test suite would be helpful to check correctness and performance. Is that already possible to implement for things like audio and video caching?

The SSE equivalent for ARM is Neon AFAIK: https://www.arm.com/why-arm/technologies/neon

PR #1782, opened 2021-10-27
traverser: catch color only jobs

ThomasWilshaw:
If a job has a color but no texture (e.g. a Solid), force the channel count to at least `VideoParams::kRGBChannelCount`.

After recent changes to `NodeTraverser::GetChannelCountFromJob()` Solid nodes caused a crash as their channel count was set to 0, causing a divide by 0 error in `Frame::set_video_params()`

### DIFF
--- a/app/node/traverser.cpp
+++ b/app/node/traverser.cpp
@@ -157,6 +157,9 @@ int NodeTraverser::GetChannelCountFromJob(const GenerateJob &job)
       }
     }
   }
+  if (max_channel_count == 0) {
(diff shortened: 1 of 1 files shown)


itsmattkc:
I think this should be implemented differently. Rather than check for a color value, change it to an `if (max_channel_count == 0)` then `max_channel_count = RGB` after the `for` loop. I think it makes sense that if no textures were found (hence 0 channel count) that we would default to RGB.

ThomasWilshaw:
Thanks, I see what you mean. Done

PR #1783, opened 2021-10-30
Copy/Paste Keyframes

### LINKED ISSUE #1349: [ANIM] Copy/paste keyframes
_Originally opened as https://github.com/sobotka/olive/issues/86_

**General TODO Summary**
Keyframes currently aren't able to be copied or pasted which is obviously not desirable from a user perspective.

**Additional Information / Notes**
We can probably use the same strategy that nodes use to copy (serializing to text by re-using their project save code) making this relatively simple to implement.

ThomasWilshaw:
Fixes #1349

### DIFF
--- a/app/node/node.cpp
+++ b/app/node/node.cpp
@@ -1941,7 +1941,31 @@ void Node::LoadImmediate(QXmlStreamReader *reader, const QString& input, int ele
     } else if (reader->name() == QStringLiteral("keyframing") && IsInputKeyframable(input)) {
       SetInputIsKeyframing(input, reader->readElementText().toInt(), element);
     } else if (reader->name() == QStringLiteral("keyframes")) {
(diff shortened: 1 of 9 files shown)


ThomasWilshaw:
@itsmattkc What do you think is the best way to make sure keyframes are pasted into the right node? We could try and figure it out based on the selected nodes or maybe just force the user to only copy/paste to/from one node at a time?

itsmattkc:
> @itsmattkc What do you think is the best way to make sure keyframes are pasted into the right node?

This is the primary reason this feature hadn't been completed yet. But I think it'll be a combination of using node IDs and prioritizing the "highlighted" node where possible.

ThomasWilshaw:
Yes I think I made the mistake of launching into this without thinking about it enough first. Okay, I'll give that a go :)

PR #1788, opened 2021-11-03
Set $APPDIR

probonopd:
You cannot rely on the `$APPDIR` variable to be populated. E.g., when the user extracts the AppImage using `--appimage-extract` and then tries to run `./squashfs-root/AppRun`, it will not be there.

Hence you need to set this variable at the beginning of your `AppRun` script:

```
APPDIR=$(readlink -f $(dirname "$0"))
```

---

Also changed the shebang to a version that works on more systems, e.g., on FreeBSD.

### DIFF
--- a/app/packaging/linux/AppRun
+++ b/app/packaging/linux/AppRun
@@ -1,4 +1,6 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+APPDIR=$(readlink -f $(dirname "$0"))
 
 # Custom AppRun that ensures the AppImage doesn't dismount before olive-crashhandler exits
 


probonopd:
The macOS build failure seems to be entirely unrelated to my change:

```
-- Configuring incomplete, errors occurred!
  Could NOT find FFMPEG (missing: FFMPEG_INCLUDE_DIRS FFMPEG_LIBRARIES avutil
See also "/Users/runner/work/olive/build/CMakeFiles/CMakeOutput.log".
  avcodec avformat avfilter swscale swresample) (Required is at least version
  "3.0")
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.21.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindFFMPEG.cmake:189 (find_package_handle_standard_args)
  CMakeLists.txt:109 (find_package)
```

itsmattkc:
I'm not familiar with the commands used to set `APPDIR` here, but it seems often they're the other way around:

- https://github.com/spesmilo/electrum/blob/master/contrib/build-linux/appimage/apprun.sh
- https://github.com/Bitcoin-ABC/ElectrumABC/blob/master/contrib/build-linux/appimage/scripts/apprun.sh

Is there a difference between these statements and yours? Also will this correctly handle paths with spaces in them (i.e. are there enough double-quotes, etc.)?

probonopd:
Good catch. `APPDIR="$(dirname "$(readlink -e "$0")")"` uses more quotes, so if that works, let's use it.

PR #1796, opened 2021-11-23
Preliminary dnxhd work

Quackdoc:
Preliminary dnxhd work to get it functioning. currently the biggest issues with this are;

1. Only works with .mov containers ffmpeg does not constrict dnxhd to spec when not using .mxf container. so this is generally bad behavior, but it should still work when using with other ffmpeg based programs. 

2. No way to control the profile it encodes too. as it is currently hardcoded. needs a gui to control it.

### DIFF
--- a/app/codec/exportformat.cpp
+++ b/app/codec/exportformat.cpp
@@ -117,7 +117,7 @@ QList<ExportCodec::Codec> ExportFormat::GetVideoCodecs(ExportFormat::Format f)
   case kFormatTIFF:
     return {ExportCodec::kCodecTIFF};
   case kFormatQuickTime:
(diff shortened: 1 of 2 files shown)


sobotka:
This is some decent ground work. Any reason why you closed the PR?

Quackdoc:
> This is some decent ground work. Any reason why you closed the PR?

I am looking to do it properly instead of a hack like this, and I am very unfamiliar with git, I plan on making a new PR when i get it actually working. apologies if this is bad etiquette, but it is far more simple for me just to close this then open up a PR with good code.

ThomasWilshaw:
You can mark PRs as drafts so people can see you're working on it but it's not finished.

sobotka:
No need to apologize. DNxHD properly integrated into Olive would be a huge win. Keep at it!

PR #1797, opened 2021-11-25
saveotio: Fix external reference available media

ThomasWilshaw:
Quick fix of OTIO saving code. External reference available medaia durataion tag were not set correctly causing OTIO converter apps to not correctly parse the generated OTIO files.

### DIFF
--- a/app/task/project/saveotio/saveotio.cpp
+++ b/app/task/project/saveotio/saveotio.cpp
@@ -135,14 +135,10 @@ OTIO::Track *SaveOTIOTask::SerializeTrack(Track *track)
     if (dynamic_cast<ClipBlock*>(block)) {
       auto otio_clip = new OTIO::Clip(block->GetLabel().toStdString());
 
-      auto rate = 0.0;
-      rational frame_rate = static_cast<ClipBlock*>(block)->connected_viewer()->GetVideoParams().frame_rate();
-      if (frame_rate.denominator() == 1) {
-        rate = frame_rate.numerator();
-      } else {
-        // Handle strange rates
-        rate = static_cast<double>(frame_rate.numerator()) /
-               static_cast<double>(frame_rate.denominator());
+      double rate = static_cast<ClipBlock*>(block)->connected_viewer()->GetVideoParams().frame_rate().toDouble();
+      if (rate == qSNaN()) {
+        // We shouldn't ever get here, but catch without crashing if we ever do
+        goto fail;
       }
 
       otio_clip->set_source_range(OTIO::TimeRange(block->in().toRationalTime(),
--- a/app/task/project/saveotio/saveotio.cpp
+++ b/app/task/project/saveotio/saveotio.cpp
@@ -136,7 +136,7 @@ OTIO::Track *SaveOTIOTask::SerializeTrack(Track *track)
       auto otio_clip = new OTIO::Clip(block->GetLabel().toStdString());
 
       double rate = static_cast<ClipBlock*>(block)->connected_viewer()->GetVideoParams().frame_rate().toDouble();
-      if (rate == qSNaN()) {
+      if (qIsNaN(rate)) {
         // We shouldn't ever get here, but catch without crashing if we ever do
         goto fail;
       }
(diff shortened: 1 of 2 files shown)


app/task/project/saveotio/saveotio.cpp
```diff
@@ -135,12 +135,33 @@ OTIO::Track *SaveOTIOTask::SerializeTrack(Track *track)
...
       auto otio_clip = new OTIO::Clip(block->GetLabel().toStdString());
 
+      auto rate = 0.0;
+      rational frame_rate = static_cast<ClipBlock*>(block)->connected_viewer()->GetVideoParams().frame_rate();
+      if (frame_rate.denominator() == 1) {
+        rate = frame_rate.numerator();
+      } else {
+        // Handle strange rates
+        rate = static_cast<double>(frame_rate.numerator()) /
+               static_cast<double>(frame_rate.denominator());
```
Simran-B, line 145: Should the denominator be checked for 0 for dividing?

itsmattkc: It might be preferable to use `rational::toDouble()` which does effectively the same thing, but does check for a 0 denominator at which point it returns NaN.

ThomasWilshaw: I think I've addressed the issue here, thanks for the review.

ThomasWilshaw:
@itsmattkc Do you think it's okay to use the sequence framerate here or is there a clean way to get the frame rate of the source media?

itsmattkc:
Sequence frame rate would make sense for a timeline wouldn't it? Usually clips are in the "sequence domain".

However, calling `connected_viewer` does not reference the sequence. It references the footage/sequence/etc that may or may not be connected as an input to the clip (it is confusing terminology).

app/task/project/saveotio/saveotio.cpp
```diff
@@ -135,12 +135,29 @@ OTIO::Track *SaveOTIOTask::SerializeTrack(Track *track)
     if (dynamic_cast<ClipBlock*>(block)) {
       auto otio_clip = new OTIO::Clip(block->GetLabel().toStdString());
 
+      double rate = static_cast<ClipBlock*>(block)->connected_viewer()->GetVideoParams().frame_rate().toDouble();
+      if (rate == qSNaN()) {
```
itsmattkc, line 139: Using the macro `qIsNaN()` is preferable to an `==` comparison

ThomasWilshaw: Thanks, done

ThomasWilshaw:
Yes I think sequnece frame rate does make most sense,, especiallyas both Olive and OTIO can easily convert between rates, I just thought I'd check.

Is there an easy way to get the sequnces frame rate then? I've not really been able to find anything.

itsmattkc:
You can get it from the track through `track->track_list()->parent()`, or just pass it through each function as a parameter, since the tracks are pulled from the sequence initially.

ThomasWilshaw:
Thanks. I think I've realised my confusion here, this PR is really about fixing `available_range` which is the length of source clip and I think it makes more sense to store that with it's own framerate. The `source_range` however is the length of a clip on the timeline and I agree that should definitely be stored in the timelines frame rate. Does that sound reasonable?

itsmattkc:
Yes I think that makes sense. In which case, you'd want to use `connected_viewer()` (usually Footage) for the `available_range` frame rate and `track->track_list()->parent()` (usually Sequence) for the `source_range` frame rate. However, you should also have a fallback check if `connected_viewer()` is null and to use the sequence in that scenario, because not all clips are going to have footage (e.g. a solid or text clip).

ThomasWilshaw:
Is the `null` check necessary as were already only running this code if the `block` has attached `Footage` nodes? Things like solids and text nodes don't need an available range anyway as they're generated right?

itsmattkc:
Ah, I guess not in that case.

ThomasWilshaw:
I think this is done on my end. This PR is really just a quick fix, I think there's still plenty of stuff to do with the saving code but I'll save that for a future PR.

itsmattkc:
Okay looks good to me

PR #1815, opened 2022-01-02
Preliminary DNxHD.

Quackdoc:
this is the better worked PR for getting DNxHD to work, as part of this I also did add preliminary x264 preset work too as they are pretty much the same function. 

just as before the DNxHD work will only work with apple quick-time .mov files. which while it isn't ideal still useful, this is necessary for proper DNxHD support regardless. 

as for x264 support, its pretty much the same code anyways. I'm not sure how to better implement it.

I tried to make the changes as un-intrusive as possible, while keeping it easy to expand upon. values are hard coded in as I cannot figure out how Qt works. however by simply changing the number for the appropriate switch statements you can change the profile and preset accordingly, so everything besides the GUI integration is at least in a working state.

unfortunately I don't have time to work on this. and won't for the foreseeable future. I hope I may be able to pick this back up, or if this can help someone pick up the task of implementing these features.

### DIFF
--- a/app/codec/exportcodec.cpp
+++ b/app/codec/exportcodec.cpp
@@ -44,6 +44,8 @@ QString ExportCodec::GetCodecName(ExportCodec::Codec c)
     return tr("PNG");
   case kCodecProRes:
     return tr("ProRes");
+    case kCodecCineform:
+    return tr("Cineform");
   case kCodecTIFF:
     return tr("TIFF");
   case kCodecMP2:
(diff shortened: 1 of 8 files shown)


Quackdoc:
Removed x264 preset support, In it's current state, DNxHD should be usable with quicktime files, with the exception of the `dnxhd` profile (needed for official 1080p and lower support) DNxHR will work with 720p and 1080p footage unofficially. 

exporting to quicktime files is should be ready for testing aside from previously mentioned errors.

PR #1818, opened 2022-01-04
Color differnece key and despill nodes

ThomasWilshaw:
Adds a color difference and despill node 

**Usage**
1. Connect plate to input then click outputm ask to view your mask. 
2. Increase "Darks" to crush the black level and increase "Brghts" to push up the whites
3. Add garbage mattes (white = remove, black =do nothing)
4. Add core mattes (white = keep, black= do nothing)
5. Add a despill node after the keyer

image

image


app/shaders/colorkey.frag
```diff
@@ -29,7 +29,8 @@ void main(void) {
     if (core_in_enabled) {
       // Force anything we want to be black
       vec4 core = texture2D(core_in, ove_texcoord);
-      mask *= core;
+      vec4 core_invert = vec4(1.0 - core.r, 1.0 - core.g, 1.0 - core.b, 1.0);
```
brimson: Can be vectorized to `vec4(1.0 - core.rgb, 1.0)`

app/shaders/despill.frag
```diff
@@ -0,0 +1,51 @@
...
+uniform int color_in;
+uniform int method_in;
+uniform bool preserve_luminance_input;
+uniform vec3 luma_coeffs;
+
+varying vec2 ove_texcoord;
+
+void main(void) {
+    vec4 tex_col = texture2D(tex_in, ove_texcoord);
+
```
brimson, line 11: Fortunately, GLSL allows the use of switch cases and converting multiply-add component math to dot products. I'll make some changes and optimizations

app/shaders/despill.frag
```diff
@@ -0,0 +1,51 @@
...
+    if(color_in == 0) { // Green screen
```
brimson, line 12: Fortunately, GLSL allows the use of switch cases and converting multiply-add component math to dot products. I'll make some changes and optimizations

itsmattkc: Is that so? I seem to remember having an issue with switch statements, perhaps on a GL ES GPU?

brimson: Yeah it's supported according to the 3.0 spec and using this code in https://shadered.org/app?fork=4Y3P9GQID0 
```glsl
#version 300 es

precision highp float;

uniform vec2 uResolution;
uniform float uTime;
const int switchyboi = 1;

out vec4 outColor;

void main()
{
	float outputcolor = 1.0;
	switch(switchyboi)
	{
		case 0:
			outputcolor = 1.0;
			break;
		case 1:
			outputcolor = 0.0;
			break;
	}
    outColor = vec4(outputcolor);
}
```
Was the issue related to GPU or the API itself?

itsmattkc: It was GPU-specific, in that it compiled fine on one GPU and not on another, but that's fairly common in my experience when going between vendors and/or desktop GL to GL ES and should probably be accounted for.

It might also be worth mentioning that we're targeting GL ES 2.0 at the moment (from an earlier attempt to improve hardware compatibility), though in all honesty we'll probably leapfrog to Vulkan over the next year (GL is showing a lot of strain with our usage, i.e. shared contexts, multithreading, float buffers, etc.)

ThomasWilshaw: I originally tried switch statements in some places but they didn't work on my setup so I changed to if/else blocks

Simran-B: The accompanying shading language of OpenGL ES 2.0 is GLSL ES 1.0(!) and it does **not support** switch statements:

> 3.7 Keywords
> …
> The following are the keywords reserved for future use. Using them will result in an error:
> …
> goto switch default

Even if the GLSL ES version number was mistaken for a GLSL version number (GLSL 1.0), it wouldn't make a difference with regards to the switch statement, because GLSL ES 1.0 is based on GLSL 1.2, which does not support switch statements either.

brimson: Thanks all, I wasn't aware that Olive reverted from 3.x to 2.0. I reverted to an if-else scheme (hope it's understandable) 👍

itsmattkc:
Not sure why GitHub didn't pick this up (maybe because I squash merged locally?) but this has been merged in 3aafb2466d6edf79bb8f42f1e5be025a52b72f0f

Issue #1781, opened 2021-10-27
[CRASH] - Move 2 clips in timeline that have cross-dissolve in them causes the crash


rpatros1979:
**Commit Hash** cf1bd65b -->

**Platform** Arch Linux -->

**Summary**
When adding 2 clips in the line that contain cross-dissolve and you move those clips couple of frames ahead, causes the crash. See attached video

**Steps to Reproduce**

1. Create a seqence
2. Create a solid on video track 0
3. Create a solid on video track 1 but it need to be ahead of video clip 0
4. Add cross-dissolve to end of clip on track 0 and at the beginning of track 1
5. Select all 4 objects on track 0 and track 1 (clip and cross-dissolve) and move it to the right
6. The crash occurs.

<details><summary><strong>Crash Report</strong></summary><pre><code>

Operating system: Linux
                  5.10.71 -1-lts #1 SMP Wed, 06 Oct 2021 14:19:52 +0000 x86_64
CPU: amd64
     family 6 model 58 stepping 9
     4 CPUs

GPU: UNKNOWN

Crash reason:  SIGSEGV /SEGV_MAPERR
Crash address: 0x30
Process uptime: 99 seconds

Thread 0 (crashed)
 0  olive-editor!olive::NodeGraph::NodeMapContainsNode(olive::Node*, olive::Node*) const [qmap.h : 68 + 0x0]
    rax = 0x0000000000000000   rdx = 0x0000000006c3e440
    rcx = 0x0000000000000018   rbx = 0x0000000003bbe2c0
    rsi = 0x0000000000000028   rdi = 0x0000000000000020
    rbp = 0x0000000003bbe2c0   rsp = 0x00007ffc7f159700
     r8 = 0x0000000003bbe2c0    r9 = 0x00007f7d9bbc3a60
    r10 = 0x0000000003bbe3e0   r11 = 0x0000000000000000
    r12 = 0x0000000006b1d090   r13 = 0x00007ffc7f159888
    r14 = 0x0000000006c1c640   r15 = 0x0000000006b669c0
    rip = 0x0000000000571588
    Found by: given as instruction pointer in context
 1  olive-editor!olive::NodeSetPositionCommand::redo() [node.cpp : 2365 + 0x5]
    rbx = 0x0000000003bbe2c0   rbp = 0x0000000003bbe2c0
    rsp = 0x00007ffc7f159730   r12 = 0x0000000006b1d090
[...]

PR #1819, opened 2022-01-05
pointer: Move attached transitions together with the clip

### LINKED ISSUE #1809: Moving a clip along with its dissolve doesn't move the dissolve
**Commit Hash**  e902507e

**Platform** Windows 10

**Summary**

I added a cross dissolve at the end of a video clip in order to fade it out. Then I select the clip as well as the dissolve and try to move everything. The dissolve just stays where it is.

Here is a recording of this behaviour:
crossDissolveNotMoving

antonhladunmobica:
Fixes #1809

Also can be a fix to #1781. Doesn't fix the cause of the crash, but should make it impossible to reproduce.

### DIFF
--- a/app/widget/timelinewidget/tool/pointer.cpp
+++ b/app/widget/timelinewidget/tool/pointer.cpp
@@ -330,6 +330,20 @@ void PointerTool::InitiateDragInternal(Block *clicked_item,
 
         // Create ghost for this block
         AddGhostFromBlock(block, trim_mode, true);
+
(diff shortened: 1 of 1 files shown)


philiplb:
Awesome! [#1809](https://github.com/olive-editor/olive/issues/1809) was quite annoying in my workflow :)

itsmattkc:
Transitions have further developed and this issue is no longer present

PR #1821, opened 2022-01-06
Preserve folder structure in visual studio

Cambloid:


ThomasWilshaw:
Could you provide some context for this PR please?

Cambloid:
without it, visual studio lists all source files at once in libolive-editor

itsmattkc:
I think I've heard of this before, is this because the Visual Studio CMake generator will put compiled objects in the same folder by default and some of our files are named the same as other files in other directories?

Cambloid:
According to the documentation:

---

New in version 3.8.

CMake will automatically detect, from <src> files paths, source groups it needs to create, to keep structure of source groups analogically to the actual files and directories structure in the project. Paths of <src> files will be cut to be relative to <root>. The command fails if the paths within src do not start with root.

---

I compiled Olive with VS 2022 and the obj files are put in the build directory which has almost the same structure than the app folder. 

Here is a screenshot on how it looks in VS:
Capture

itsmattkc:
Ah, so this is a flag specifically Visual Studio the IDE?

Also it looks like you've committed extra things that aren't meant to be part of this PR.

Cambloid:
It is a command for the IDE project generation so not just Visual Studio.

Fixed it, moved the changes to a separate branch.

PR #1828, opened 2022-01-12
Numbering Nodes

GUY-BBS:
In order to demonstrate Olive's node features, I created the animated strap shown in the two videos below

https://user-images.githubusercontent.com/49677111/149140451-fb1584c3-af9a-4169-8386-02500edf8468.mp4

https://user-images.githubusercontent.com/49677111/149140463-fb7d9f8c-163d-4487-9aea-6e04c356824c.mp4

The creation of the node structure and setting the parameters took more time than I expected because of the difficulty of identifying which of the nodes I was addressing in the parameter editor window.
Could the nodes be numbered as in the graphic mock-up below?
.Also if each parameter window could be minimised, (as suggested by the arrowhead beside the title header) it would allow you to open the relevant section
Numbering Nodes
.


ThomasWilshaw:
Hi, thanks for the demos and feedback, it's good to see people doing interesting things with nodes. 

You can already label nodes anything you like (right click on the node and select label). Properties windows should collapse when you select the title bar, if there not that's a bug, please file a issue in the issue tracker. 

On that note, issues like this should be opened as Issues, not Pull Requests, unless you are submitting code to solve the problem.

Thanks

GUY-BBS:
Many thanks for the info on labelling nodes. I will analyse why I am having problems clicking in the title bar on a couple of different computers before I submit it as a bug,

PR #1831, opened 2022-01-14
Text typewriter feature

franga2000:
Hi! I developed this feature out of necessity and figured it might be useful to others too. I enables the "typewriter" effect on text nodes, where the text is "typed in" one character at a time. 

Would you accept this feature? If so, what do I need to improve before it's ready? Some thoughts:

 - [ ] Better name for it? Resolve calls it "Write On", After Effects has a "Typewriter" effect
 - [ ] Support for HTML text (probably possible with `QTextCursor`)
 - Maybe it would make sense to scrap this implementation and add a "Substring" node that could be plugged into a Text node and keyframed? *I think that could be possible, but I'm completely new to the project so I'm not sure how flexible the node system is...*

Demo:   

https://user-images.githubusercontent.com/3891092/149542312-2f7a8d4c-5f16-44f5-8ceb-4d2ad8620f59.mp4

### DIFF
--- a/app/node/generator/text/text.cpp
+++ b/app/node/generator/text/text.cpp
@@ -20,6 +20,8 @@
 
 #include "text.h"
 
+#include "common/clamp.h"
+#include "widget/slider/floatslider.h"
 #include <QAbstractTextDocumentLayout>
 #include <QDateTime>
 #include <QTextDocument>
@@ -47,6 +49,7 @@ const QString TextGenerator::kHtmlInput = QStringLiteral("html_in");
 const QString TextGenerator::kVAlignInput = QStringLiteral("valign_in");
 const QString TextGenerator::kFontInput = QStringLiteral("font_in");
 const QString TextGenerator::kFontSizeInput = QStringLiteral("font_size_in");
+const QString TextGenerator::kTextTypewriterInput = QStringLiteral("text_typewriter_in");
 
 TextGenerator::TextGenerator()
 {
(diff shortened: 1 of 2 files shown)


itsmattkc:
A "typewriter effect" is obviously useful, however I wouldn't accept it in the way you've done it (attaching it to the text node). The nodes are like building blocks and each one should have a single clearly defined purpose.

Instead, I'd be inclined to make a "substring" node that would function basically identically to the substring function of any programming language. It could take an input string, have inputs for start/length, and then those could be keyframed (or have "time" plugged into it). That would provide much more versatility and flexibility, and you could even group it all together into one "typewriter" node with the new grouping feature if you want to keep it all tidy and package it as a single node (that can be copied/pasted into other projects or to other users).

franga2000:
Yeah, I figured that would make more sense. I already have a substring node mostly implemented, so I'll push it here once finished

ThomasWilshaw:
Olive is about to undergo some largish code refactoring and I wanted to check on the status of old PRs before we do that. Do you olan to update/continue with the PR? Thanks

franga2000:
The status is...I forgot about it, sorry :sweat_smile:

I doubt I still have my unfinished substring node code and I don't have time to finish this atm. I'll try to remember to come back and re-do it some time in the summer when I have a bit more time.

TheTechRobo:
Thought I'd chime in with my 2cents: Maybe for more complicated effects, a "preset" menu could be used that has some basic, commonly-used effects. (For something like this that could be done with a substring node, it's probably not necessary, but might be nice for more complicated effects.)

PR #1845, opened 2022-01-24
Dilate/Erode node

ThomasWilshaw:
Adds a dilate erode node to the filter category. Mainly useful for masks and other grey scale images. 

Only controversial issue is that the Gaussian erode does an inversion via `1.0 - col.rgb` which isn't really okay colour management wise. Is there a better way to do this? It may be okay in this case as this node will only really be used for masks or it may be better to wait until we have implemented an OCIO node so this operation can be done in screen space if required.

### DIFF
--- a/app/node/factory.cpp
+++ b/app/node/factory.cpp
@@ -39,6 +39,7 @@
 #include "generator/text/text.h"
 #include "generator/text/textlegacy.h"
 #include "filter/blur/blur.h"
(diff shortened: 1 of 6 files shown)


ThomasWilshaw:
All that is left here is to optimise the shader to run in two passes, horizontal and vertical. This would change the run time from O(n^2) to O(n)

ThomasWilshaw:
I think this is complete now @itsmattkc

PR #1847, opened 2022-01-24
Ensure jack client is named "Olive" on the first initialization and make use of the pa_jack header dependant on its availability.

parona-source:
* Switch around Pa_Initialize and PaJack_SetClientName so that the jack
  client is named "Olive" on the first initialization and not on
  subsequent initializations when refreshing devices.
* Make use of pa_jack.h dependant on availability of the header and not
  on the usage of the linux platform.
  
  This is what I was going about on the discord server.


CMakeLists.txt
```diff
@@ -141,6 +141,13 @@ list(APPEND OLIVE_LIBRARIES
 
 # Link PortAudio
 find_package(PortAudio REQUIRED)
+set(CMAKE_REQUIRED_INCLUDES ${PORTAUDIO_INCLUDE_DIRS})
+include(CheckIncludeFileCXX)
+check_include_file_cxx( "pa_jack.h" PA_HAS_JACK)
+if (PA_HAS_JACK)
+  add_compile_definitions(PA_HAS_JACK)
```
itsmattkc, line 148: `list(APPEND OLIVE_DEFINITIONS -DPA_HAS_JACK)` would be preferable to the global `add_compile_definitions`

CMakeLists.txt
```diff
@@ -141,6 +141,13 @@ list(APPEND OLIVE_LIBRARIES
 
 # Link PortAudio
 find_package(PortAudio REQUIRED)
+set(CMAKE_REQUIRED_INCLUDES ${PORTAUDIO_INCLUDE_DIRS})
```
itsmattkc, line 144: What does this line do?

parona-source: https://cmake.org/cmake/help/v3.13/module/CheckIncludeFileCXX.html Its there to ensure that CheckIncludeFileCXX checks the correct include dir which the bundled FindPortAudio.cmake should point to.

CMakeLists.txt
parona-source: True, I didnt notice due tunnel vision from intially trying to get it to work in app/audio/CMakeLists.txt instead of doing it in the root CMakeLists.txt. Ill change this and then rebase.

CMakeLists.txt
itsmattkc: No, the `set(CMAKE_REQUIRED_INCLUDES ${PORTAUDIO_INCLUDE_DIRS})` line

itsmattkc: Ah I see

PR #1848, opened 2022-01-24
Noise node

ThomasWilshaw:
Adds a simple noise node based on the 0.1 shader code.


app/node/generator/noise/noise.cpp
```diff
@@ -0,0 +1,82 @@
...
+}
+
+QString NoiseGeneratorNode::Name() const
+{
+  return tr("Noise");
+}
+
+QString NoiseGeneratorNode::id() const
+{
+  return QStringLiteral("org.olivevideoeditor.Olive.blue");
```
itsmattkc, line 47: `blue`?

ThomasWilshaw: I've no idea :')

ThomasWilshaw: Done

PR #1850, opened 2022-01-25
Add Flip/Flop node

ThomasWilshaw:
Title says it all.


itsmattkc:
Looks good, but any chance we can just call it "Flip" lol

ThomasWilshaw:
Sure, is that internally as well as in the UI?

itsmattkc:
Preferably both, but mainly UI

ThomasWilshaw:
Done

PR #1851, opened 2022-01-25
Corner pin

ThomasWilshaw:


app/node/distort/cornerpin/cornerpindistortnode.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+  ShaderJob job;
+  job.InsertValue(value);
+  job.SetAlphaChannelRequired(GenerateJob::kAlphaForceOn);
+  job.InsertValue(QStringLiteral("resolution_in"), NodeValue(NodeValue::kVec2, globals.resolution(), this));
+
+  // Convert slider values to their pixel values and then convert to clip space (-1.0 ... 1.0) for overriding the
+  // vertex coordinates.
+  const QVector2D &resolution = globals.resolution();
+  QVector2D top_left = QVector2D(ValueToPixel(0, value, resolution)) / (resolution / 2.0) - QVector2D(1.0, 1.0);
```
itsmattkc, line 66: Might be good to cache `resolution / 2.0` instead of calculating it each time, also probably preferable to multiply by 0.5 instead (multiplication is generally faster than division)

app/node/distort/cornerpin/cornerpindistortnode.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+  p->setPen(QPen(Qt::white, 0));
+
+  QPointF top_left = ValueToPixel(0, row, resolution);
+  QPointF top_right = ValueToPixel(1, row, resolution);
+  QPointF bottom_right = ValueToPixel(2, row, resolution);
+  QPointF bottom_left = ValueToPixel(3, row, resolution);
+
+  // Add the correct offset to each slider
+  SetInputProperty(kTopLeftInput, QStringLiteral("offset"), QVector2D(0.0, 0.0));
```
itsmattkc, line 142: It might be good if `ValueToPixel` could be used to calculate the offsets too. It would mean ValueToPixel would have to take a QVector2D instead of a NodeValueRow, but you could have an additional override that pulls the QVector2D from the row so both usages are valid. e.g:

```
QPointF CornerPinDistortNode::ValueToPixel(int value, const NodeValueRow& row, const QVector2D &resolution) const
{
  QString column;
  switch (value) {
  case 0:
    column = kTopLeftInput;
    break;
  etc. etc.
  }

  return ValueToPixel(value, row[column].data().value<QVector2D>(), resolution);
}
```

app/node/distort/cornerpin/cornerpindistortnode.cpp
```diff
@@ -0,0 +1,237 @@
...
+    gizmo_active[i] = gizmo_resize_handle_[i].contains(p);
+
+    if (gizmo_active[i]) {
+      gizmo_drag_start_ = p;
+      gizmo_res_ = globals.resolution();
+      gizmo_drag_ = i;
+
+      switch (i) {
+        case 0:
+          gizmo_start_ = row[kTopLeftInput].data();
```
itsmattkc, line 177: NodeInputDragger now has a `GetStartValue` function which can be used rather than having to cache the start value in `gizmo_start_`. Note that GetStartValue will be a double rather than a QVector2D since it's only one track's value.

app/render/job/shaderjob.h
```diff
@@ -87,6 +88,19 @@ class ShaderJob : public GenerateJob {
     interpolation_.insert(id, interp);
   }
 
+  void SetVertexCoordinates(QVector<float> vertex_coords)
+  {
+    vertex_overrides_.clear();  
+    foreach (float value, vertex_coords) {
+        vertex_overrides_.append(value);
+    }
```
itsmattkc, line 96: This could simply be `vertex_overrides_ = vertex_coords;` and would have the same effect. Also changing the parameter to `const QVector<float> &vertex_coords` is preferable to prevent deep copies.

app/render/job/shaderjob.h
```diff
@@ -87,6 +88,19 @@ class ShaderJob : public GenerateJob {
...
+  }
+
+  QVector<float> GetVertexCoordinates()
```
itsmattkc, line 99: This too should be changed to `const QVector<float> &GetVertexCoordinates()` to prevent deep copies.

itsmattkc (changes requested):

app/node/distort/cornerpin/cornerpindistortnode.cpp
ThomasWilshaw: Sorry I'm not quite sure what you mean here.

PR #1853, opened 2022-02-01
Cineform support

Quackdoc:
Adds in initial cineform support, only adding for quicktime right now, but I think it would work in matroska too?

It defaults to filmscan 3, which is probably not what most people want (even in the SDK it's listed as overkill [see here](https://github.com/gopro/cineform-sdk/blob/9785875ba0555f92889cac0ded94304ec49ced4c/Common/CFHDTypes.h#L208)) but it's what ffmpeg defaults to anyways, however in terms of testing, it works perfectly fine, and we can always configure it when codec configuration settings eventually get added, which is needed for dnxhd support anyways.

### DIFF
--- a/app/codec/exportcodec.cpp
+++ b/app/codec/exportcodec.cpp
@@ -44,6 +44,8 @@ QString ExportCodec::GetCodecName(ExportCodec::Codec c)
     return tr("PNG");
   case kCodecProRes:
     return tr("ProRes");
+    case kCodecCineform:
+    return tr("Cineform");
   case kCodecTIFF:
     return tr("TIFF");
   case kCodecMP2:
@@ -79,6 +81,7 @@ bool ExportCodec::IsCodecAStillImage(ExportCodec::Codec c)
   case kCodecH264rgb:
   case kCodecH265:
   case kCodecProRes:
+  case kCodecCineform:
   case kCodecMP2:
   case kCodecMP3:
   case kCodecAAC:
--- a/app/codec/exportcodec.cpp
+++ b/app/codec/exportcodec.cpp
@@ -44,8 +44,6 @@ QString ExportCodec::GetCodecName(ExportCodec::Codec c)
     return tr("PNG");
   case kCodecProRes:
     return tr("ProRes");
-    case kCodecCineform:
-    return tr("Cineform");
   case kCodecTIFF:
     return tr("TIFF");
   case kCodecMP2:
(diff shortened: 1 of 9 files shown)


itsmattkc:
> we can always configure it when codec configuration settings eventually get added

We already have a system of codec configuration, the snag is that every codec requires different settings/considerations, and therefore its own specific UI widget. Currently we only have one for H.264 (known as `H264Section`) and still images (known as `ImageSection`), but it should be very much possible to write a `CineformSection` to configure the settings requried for it too.

Quackdoc:
I did see that, but is that the route we want to go for these? for dnxhd, and cineform, there is really only one parameter we care about tweaking, and for prores, maybe 2 parameters? it seems a little overkill, maybe something like an `IntermediateSection` to encompass all three? 

but I don't know much about QT itself, so I'm not sure how it would look.

itsmattkc:
I think it makes sense to encapsulate and abstract them like that, even if it is just one parameter each. I understand thinking it's overkill, but I think readability and maintainability here is preferable to having as minimal code as possible.

Quackdoc:
That's fair, I'm not really sure how it should work, as I said, I know very little about QT it'self and all my attempts to learn have proved... non fruitful. however all we really need to do is pass it what presets we need, so `av_opt_set()` or `av_dict_set()` are pretty standard ways of doing so.

itsmattkc:
If you can tell me the setting to pass to `av_opt_set`/`av_dict_set` I can probably add it quickly to this PR

Quackdoc:
Here are the inputs, the `quality` argument takes either the integer `10`, or the name of the quality `medium` as the input. a reasonable default would probably be medium or high.

```
Encoder cfhd [GoPro CineForm HD]:
    General capabilities: threads 
    Threading capabilities: frame
    Supported pixel formats: yuv422p10le gbrp12le gbrap12le
cfhd AVOptions:
  -quality           <int>        E..V....... set quality (from 0 to 12) (default film3+)
     film3+          0            E..V.......
     film3           1            E..V.......
     film2+          2            E..V.......
     film2           3            E..V.......
     film1.5         4            E..V.......
     film1+          5            E..V.......
     film1           6            E..V.......
     high+           7            E..V.......
     high            8            E..V.......
     medium+         9            E..V.......
     medium          10           E..V.......
     low+            11           E..V.......
     low             12           E..V.......


```

itsmattkc:
Okay I've added a CineformSection, let me know if it functions as expected.

Quackdoc:
The resulting files look about right, other then a strange cosmetic issue on my end it looks fine, there is a bit of a strange gap, here is a pic to showcase it, other than that, like I said, the resulting videos bitrates are about right and line up with manually doing it! Thanks for the help, I think I can also use your commit is a guideline for DNxHD (whenever I find out that pesky framerate issue) as well as connect the nobs for prores.

https://files.catbox.moe/qss239.png

itsmattkc:
Ah yes, I saw that but forgot to fix. Should be good now. If the files it produces are correct, I think this should be good to merge.

PR #1856, opened 2022-02-02
Add support for x264 and x265 presets

Quackdoc:
Adding presets for x264 and 265 encoders. the output lines up with the expected values for each preset on both x265 and x264.

[Values for presets for x265](https://x265.readthedocs.io/en/stable/presets.html)
[Values for presets for x264](https://dev.beandog.org/x264_preset_reference.html)

### DIFF
--- a/app/dialog/export/codec/h264section.cpp
+++ b/app/dialog/export/codec/h264section.cpp
@@ -42,6 +42,26 @@ H264Section::H264Section(int default_crf, QWidget *parent) :
   layout->setMargin(0);
 
   int row = 0;
+  layout->addWidget(new QLabel(tr("Preset:")), row, 0);
+
+  preset_combobox_ = new QComboBox();
+
+  preset_combobox_->addItem(tr("Ultrafast"));
+  preset_combobox_->addItem(tr("Superfast"));
+  preset_combobox_->addItem(tr("Veryfast"));
+  preset_combobox_->addItem(tr("Faster"));
+  preset_combobox_->addItem(tr("Fast"));
+  preset_combobox_->addItem(tr("Medium"));
+  preset_combobox_->addItem(tr("Slow"));
+  preset_combobox_->addItem(tr("Slower"));
+  preset_combobox_->addItem(tr("Veryslow"));
+  
+  //Default to "medium"
+  preset_combobox_->setCurrentIndex(5);
+
+  layout->addWidget(preset_combobox_, row, 1);
+
+  row++;
 
   layout->addWidget(new QLabel(tr("Compression Method:")), row, 0);
 
@@ -110,6 +130,8 @@ void H264Section::AddOpts(EncodingParams *params)
     params->set_video_buffer_size(2000000);
 
   }
+  
+  params->set_video_option(QStringLiteral("preset"), QString::number(preset_combobox_->currentIndex()));
 }
 
 H264CRFSection::H264CRFSection(int default_crf, QWidget *parent) :
--- a/app/dialog/export/codec/h264section.cpp
+++ b/app/dialog/export/codec/h264section.cpp
@@ -42,19 +42,25 @@ H264Section::H264Section(int default_crf, QWidget *parent) :
   layout->setMargin(0);
 
   int row = 0;
-  layout->addWidget(new QLabel(tr("Preset:")), row, 0);
+  layout->addWidget(new QLabel(tr("Encode Speed:")), row, 0);
 
+
+  char presettooltip[] = "<head/><body>This setting allows you to tweak the ratio of export speed to compression quality. <br /> <br />"\
+    " If using Constant Rate Factor, slower speeds will result in smaller file sizes for the same quaity. <br /> <br />"\
+    " If using Target Bit Rate or Target File Size, slower speeds will result in higher quality for the same bitrate/filesize. <br /> <br />"\
+    " This setting is equivalent to the `preset` setting in libx264.<head/>";
   preset_combobox_ = new QComboBox();
+  preset_combobox_->setToolTip(presettooltip);
 
-  preset_combobox_->addItem(tr("Ultrafast"));
-  preset_combobox_->addItem(tr("Superfast"));
-  preset_combobox_->addItem(tr("Veryfast"));
+  preset_combobox_->addItem(tr("Ultra Fast"));
+  preset_combobox_->addItem(tr("Super Fast"));
+  preset_combobox_->addItem(tr("Very Fast"));
   preset_combobox_->addItem(tr("Faster"));
   preset_combobox_->addItem(tr("Fast"));
   preset_combobox_->addItem(tr("Medium"));
   preset_combobox_->addItem(tr("Slow"));
   preset_combobox_->addItem(tr("Slower"));
-  preset_combobox_->addItem(tr("Veryslow"));
+  preset_combobox_->addItem(tr("Very Slow"));
   
   //Default to "medium"
   preset_combobox_->setCurrentIndex(5);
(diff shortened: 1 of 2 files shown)


app/dialog/export/codec/h264section.cpp
```diff
@@ -42,6 +42,26 @@ H264Section::H264Section(int default_crf, QWidget *parent) :
...
+
+  preset_combobox_->addItem(tr("Ultrafast"));
+  preset_combobox_->addItem(tr("Superfast"));
+  preset_combobox_->addItem(tr("Veryfast"));
+  preset_combobox_->addItem(tr("Faster"));
+  preset_combobox_->addItem(tr("Fast"));
+  preset_combobox_->addItem(tr("Medium"));
+  preset_combobox_->addItem(tr("Slow"));
+  preset_combobox_->addItem(tr("Slower"));
+  preset_combobox_->addItem(tr("Veryslow"));
```
itsmattkc, line 57: In this scenario, it might make sense to retain the naming convention of the original presets (all lowercase). Either that or convert them to fully human names like "Very Slow" and "Super Fast", but I feel the x264/5 presets are so entrenched in their original naming that using their original names makes the most sense.

sobotka: Who is the audience of Olive? LibX26x wizards or editors?

Quackdoc: @itsmattkc that's fair, I'll update the PR to reflect this, I also left out placebo, as I don't think anyone would ever want to use it, but I can add it for the sake of completion, as the preset does exist, If you think that would be best

itsmattkc: @sobotka It's an interesting question here, my thinking was that this setting in particular was esoteric and libx264-specific enough that it may in fact be the domain of libx264 wizards (or at least people well-versed in encoding) who would prefer the internal name. But I definitely see the argument to humanizing them further as I mentioned.

Quackdoc: What would be a good way of humanizing it? Calling the "preset" part, "Encode Speed"? We are using CRF by default, so we could use "Encode Speed:" "Very Fast". and I think the vast majority of people, this would work fine as the most  important thing the majority of people will need to know is, "the slower the speed, smaller the file, and longer encode times". (to an extent of course)

I think anyone who this doesn't apply for, would either likely be better served by the intermediate codecs, or would at least have enough knowledge to know what the presets do.

itsmattkc: Yeah, a tooltip could easily explain that (and its relationship to the libx264 "preset" setting) if that's the route we ended up going.

Quackdoc: [This](https://cdn.discordapp.com/attachments/325506936269307905/938659336564113428/unknown.png) is how it looks when humanized.

and [this](https://cdn.discordapp.com/attachments/325506936269307905/938659964438851604/unknown.png) is how it looks when left as it would be otherwise

itsmattkc: Honestly I do kind of like it humanized, more than I expected. What do you think?

Quackdoc: I am inclined to agree. I think it looks better, even if my inner nerd is a little dissatisfied, It is a lot more pleasant too look at, and I think it conveys the effect well enough. I'm don't know how to implement tool tips yet if we do want to use them, but if this is good, I will update the PR accordingly.

sobotka: Maybe the tooltip idea is a good one to explain the nerd laden x264 values?

itsmattkc: That's what I was thinking, the tooltip would explain both the quality/speed concept to the layman, and also explain how it relates to the x264 values for encoding experts

@Quackdoc Tooltip can be set with `preset_combobox_->setToolTip(tr(""));` in the constructor.

Quackdoc: Here are some ideas for what to put. I think anyone of them is fine, I personally like to keep things short and concise, but the second and third are also nice as it points them to the appropriate place to further their research. I think anymore description would be redundant, it's good enough to point them in the right direction, but no need to hold their hand while we do it.

1. "The slower the speed, the smaller the file." [img](https://media.discordapp.net/attachments/325506936269307905/938677520730050600/unknown.png?width=960&height=206)
2. "The slower the speed, the smaller the file. Each setting corresponds with the equivalent ffmpeg preset." [img](https://media.discordapp.net/attachments/325506936269307905/938677156429582346/unknown.png?width=960&height=202)
3. "The slower the speed, the smaller the file. Each setting corresponds with the equivalent libx264 preset." [img](https://media.discordapp.net/attachments/325506936269307905/938678239658934332/unknown.png?width=960&height=172)

itsmattkc: I'd probably be even more descriptive. Something like:

> This setting allows you to tweak the ratio of export speed to compression quality. If using CRF, slower speeds will result in smaller file sizes for the same quaity. If using target bit rate or target file size, slower speeds will result in higher quality for the same bit rate/size.
>
>This setting is equivalent to the "preset" setting in libx264.

Quackdoc: Here is the updated [tooltip](https://cdn.discordapp.com/attachments/325506936269307905/938857566828650556/unknown.png).

itsmattkc: Seems good, maybe expand "CRF" to "constant rate factor" since that's what it's called in the UI. Also there seems to be a newline at the end? But yeah otherwise that seems good

app/dialog/export/codec/h264section.cpp
```diff
@@ -42,6 +42,32 @@ H264Section::H264Section(int default_crf, QWidget *parent) :
   layout->setMargin(0);
 
   int row = 0;
+  layout->addWidget(new QLabel(tr("Encode Speed:")), row, 0);
+
+
+  char presettooltip[] = "<head/><body>This setting allows you to tweak the ratio of export speed to compression quality. <br /> <br />"\
```
itsmattkc, line 48: This is the wrong way to do this. You should be using `tr("string")` instead of declaring a char array. There's also no need for HTML tags. Just use `\n` for newlines instead.

Quackdoc: Before I mess something else up, code should look like this then? It compiles and looks fine, or are spaces preferred between then newlines?

```
  preset_combobox_->setToolTip(tr("This setting allows you to tweak the ratio of export speed to compression quality. \n\n"\
    " If using Constant Rate Factor, slower speeds will result in smaller file sizes for the same quaity. \n\n"\
    " If using Target Bit Rate or Target File Size, slower speeds will result in higher quality for the same bitrate/filesize. \n\n"\
    " This setting is equivalent to the `preset` setting in libx264."));
```

itsmattkc: You don't need the spaces in each line, you can just have `"If using Constant Rate Factor` and end with `for the same quality.\n\n"` (also a typo in "quality"). Also you shouldn't need a `\` at the end of the line, the compiler should just combine adjacent string literals.

Quackdoc: thanks, I've updated the PR. it's hard to juggle languages and different coding styles, so it helps alot

itsmattkc: Indeed, especially when it comes to strings in C++.

Anyways, this looks good to me. I'll go ahead and merge once the CI is done.

PR #1866, opened 2022-02-26
Fix the error "‘AV_CH_LAYOUT_STEREO’ was not declared in this scope" during build

tecnotercio:
The build of Olive under Arch Linux (package "olive-git") crashes in the file "app/dialog/sequence/sequencedialogpresettab.cpp".

GCC tries to look for the directive "AV_CH_LAYOUT_STEREO", but cannot find it, because the source code has no reference to the header "render/audioparams.h".

The following link contains more details about the error: https://aur.archlinux.org/packages/olive-git#comment-852760


itsmattkc:
Thanks, I thought I'd gotten all of the new FFmpeg headers but I must have missed this one.

PR #1875, opened 2022-03-08
Initial OCIO node implementation

ThomasWilshaw:
This PR adds the functionality to include OCIO shaders into a node and implements a demonstration OCIO Display Transform node. Whilst this might not be the most useful node or even one we want it seemed like a good test case. The main questions I have are:
- Is this the right way to implement this?
- Should I pull some of the functions out into a `OCIOBaseNode`?
- Is there a better way to acces the values sleected in a ComboBox?

I'm sure there is plenty of scope to clean this up too.


ThomasWilshaw:
Also I'm no really sure why the LInux/Mac are not building

AngeloFrancabandiera:
> Also I'm not really sure why Linux/Mac are not building

I have tried to build your branch on my Ubuntu Linux desktop with gcc 9.4.0 and it builds fine. In particular, the line of code that generates the error in the report does not generate even a warning. However the report refers to gcc 9.3.1.
I noticed that, in the code I have cloned, function `NodeTraverser::ProcessShader` takes a constant reference to `ShaderJob`, while the report indicates a non constant reference:

```
295 | TexturePtr NodeTraverser::ProcessShader(const Node *node, const TimeRange &range, ShaderJob &job)
    |                                                                                   ~~~~~~~~~~~^~~
```

app/node/keying/chromakey/chromakey.cpp
```diff
@@ -87,6 +87,7 @@ void ChromaKeyNode::Value(const NodeValueRow &value, const NodeGlobals &globals,
     job.SetColorProcessor(processor());
     job.SetInputTexture(value[kTextureInput].data().value<TexturePtr>());
     job.SetNeedsCustomShader(this);
+    job.SetFunctionName(QString("SceneLinearToCIEXYZ_d65"));
```
itsmattkc, line 90: `QStringLiteral()` is preferable here for performance reasons

app/render/job/colortransformjob.h
```diff
@@ -71,6 +71,9 @@ class ColorTransformJob : public AcceleratedJob
   const QMatrix4x4 &GetCropMatrix() const { return crop_matrix_; }
   void SetCropMatrix(const QMatrix4x4 &m) { crop_matrix_ = m; }
 
+  const QString &FunctionName() const { return function_name_; }
```
itsmattkc, line 74: Should be `GetFunctionName`

app/render/renderer.cpp
```diff
@@ -106,10 +106,15 @@ bool Renderer::GetColorContext(const ColorTransformJob &color_job, Renderer::Col
...
+    QString ocio_func_name;
+    if (color_job.FunctionName().isEmpty()) {
+      ocio_func_name = "OCIODisplay";
+    } else {
+      ocio_func_name = color_job.FunctionName();
+    }
     auto shader_desc = OCIO::GpuShaderDesc::CreateShaderDesc();
     shader_desc->setLanguage(OCIO::GPU_LANGUAGE_GLSL_ES_3_0);
-    shader_desc->setFunctionName(ocio_func_name);
+    shader_desc->setFunctionName(ocio_func_name.toStdString().c_str());
```
itsmattkc, line 117: Can do this faster with `ocio_func_name.toUtf8()`

PR #1882, opened 2022-03-17
8 new styles and "dark" and "light" updates

pgilfernandez:
Since a while I have been working on Olive styles:
- updating "dark" and "light" ones so that they feel more natural and comfortable when using Olive for a long period of time
- I have updated the styles from "Olive dark" and " Olive light" to just "Dark" and "Light", I think it makes sense
- creating new styles: I'm unsure if all of them should be in master, probably I would drop 2-4 of them, but I wanted to open a discussion about it so I finally added the ones that I feel are the best ones

Following up a quick screenshot of all themes (including actual ones) so that makes it easy to discuss:
### Dark (actual)
dark

### Dark (updated)
dark2

### Darker
darker

### Grey
grey

### Light (actual)
light

### Light (updated)
light2

### Olive
olive

### Resolve
resolve

### Seasons - fall
seasons_fall

### Seasons - spring
seasons_spring

### Seasons - summer
seasons_summer

### Seasons - winter
seasons_winter

Cheers


mamarilmanson:
For the summer... just an idea to use this image as a color reference... (or are we imagining different kinds of summer?) 

(just an idea...)

:)

image

GUY-BBS:
I would suggest a trivial change to Olive's GUI, which I think would significantly improve Olive's appearance. Could the folder icons be buff coloured and the clip icons be changed to thumbnails. I have uploaded two images to show the difference.
File Folder GFX A
File Folder GFX B

ThomasWilshaw:

Simran-B:
I appreciate the effort, but I'm against merging this work into the main repository.
- This PR adds over 50 MB of data, bloating the repo forever. Each CI run has to clone the full repo.
- Committing generated files is a questionable practice. If the changes are infrequent and small in size, it's not that big of a deal, but I would generally avoid it.
- The themes don't add anything functional IMO, in the sense that they merely cater to different tastes but don't add any value in terms of accessibility, for example (high-contrast, Protanopia, Deuteranopia, Tritanopia, ...)

I would rather like to see the theming support to be extended on the code side so that additional themes don't require to compile Olive. Such themes can then be added to a community repo, easily downloaded and installed, without creating a maintenance burden.

PR #1888, opened 2022-03-31
German translation improvements (WIP)

mara004:
Resumed improving the German translation.
Important note to self: I need to look through all key translations again because lupdate's heuristics did awkward stuff with these (randomly filled in different characters as default).

### DIFF
--- a/app/ts/de_DE.ts
+++ b/app/ts/de_DE.ts
@@ -63,7 +63,7 @@
     <message>
         <location filename="../config/config.cpp" line="229"/>
         <source>Failed to save application settings. The application may lack write permissions for this location.</source>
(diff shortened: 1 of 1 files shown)


Suk-RAM:
Hi mara004,

do you know when your German translation improvements get included into the daily releases? Right now these releases show some English terms, because the German counterparts are missing.

Thanks,

Suk-RAM

mara004:
@Suk-RAM Thank you for asking. I'm afraid I'm currently busy with my Python projects and don't know if/when I will resume this work, so I'd be glad if someone else could pick up the German translation. (My system is currently configured to en_US, and I'm using Shotcut or Cinelerra rather than Olive these days, which means I haven't got much personal interest in this anymore.)

mara004:
Closing because of my inactivity.

PR #1898, opened 2022-04-20
Add ID property when copying node & timeline data to avoid mixing

### LINKED ISSUE #1897: [CRASH] Copy/paste doesn't distinguish between timeline and node graph
**Commit Hash** 
3661599213e47404ea0d5021e2bbfa16bcc80790

**Platform** 
All

**Summary**
If you copy nodes from the node graph, but paste while another panel (e.g. the timeline) is focused, it will try to paste those nodes into the timeline and crash since it isn't valid timeline data. Presumably there will be issues vice versa too.

**Steps to Reproduce**

1. Copy any nodes from the node panel
2. Focus the timeline by clicking anywhere on it
3. Paste from clipboard
4. Crash will occur

aeroc7:
Fixes #1897


itsmattkc:
I decided to implement this a little differently in fb53931f6816bfb1a4ee536fcf6dcf0c2f5cb1a4 by wrapping everything in appropriate tags. This way, the code can skip the whole block of data if it's not the right receiver, rather than still having to load everything and then read its position data. Thanks for the PR though, I appreciate your time

PR #1905, opened 2022-04-22
Add sanitizer options to build system

aeroc7:
This is for easy-toggling of the sanitizer options. Options are as follows:

`cmake -DENABLE_SANITIZER_ADDRESS=TRUE ..`
`cmake -DENABLE_SANITIZER_LEAK=TRUE ..`
`cmake -DENABLE_SANITIZER_UNDEFINED_BEHAVIOR=TRUE ..`
`cmake -DENABLE_SANITIZER_THREAD=TRUE ..`
`cmake -DENABLE_SANITIZER_MEMORY=TRUE ..`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,12 @@ set(CMAKE_AUTOMOC ON)
 set(CMAKE_AUTOUIC ON)
 set(CMAKE_AUTORCC ON)
 
+# Sanitizers
+add_library(olive-sanitizers INTERFACE)
+include(cmake/Sanitizers.cmake)
+enable_sanitizers(olive-sanitizers)
+list(APPEND OLIVE_LIBRARIES olive-sanitizers)
+
 # Set compiler options
 if(MSVC)
   set(OLIVE_COMPILE_OPTIONS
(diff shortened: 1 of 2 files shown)


ThomasWilshaw:
MSVC does have a address sanitizer available (https://docs.microsoft.com/en-us/cpp/sanitizers/asan?view=msvc-160). Ot might be good to add that to this PR too for completeness.

sobotka:
Minor detail; would be prudent to drop the lingo around “sane”. Flip for validate.

ThomasWilshaw:
> Minor detail; would be prudent to drop the lingo around “sane”. Flip for validate.

Please correct me if I'm wrong but I thought this was sanitize as in clean, not as in "sanity check"  (which I agree should be dropped)

elsandosgrande:
> > Minor detail; would be prudent to drop the lingo around “sane”. Flip for validate.
> 
> Please correct me if I'm wrong but I thought this was sanitize as in clean, not as in "sanity check" (which I agree should be dropped)

Yeah, it stems from “sanitation” from what I can tell.

aeroc7:
> MSVC does have a address sanitizer available (https://docs.microsoft.com/en-us/cpp/sanitizers/asan?view=msvc-160). It might be good to add that to this PR too for completeness.

Added!

Issue #1899, opened 2022-04-20
[CRASH] Shutdown is not clean when closing the window while cache is occuring


aeroc7:
**Commit Hash** 
3661599

**Platform** 
All

**Summary**
If a pre-cache is occurring and the window is closed, a seg fault will occur. This is presumably due to some part of the cache system not receiving a proper signal to shutdown, so while other parts of the program are being closed down, the cache system continues on by itself.

**Steps to Reproduce**

1. Import a clip & pre-cache it
2. Close the window (save or don't save, it doesn't matter)
3. Seg fault will occur

<details><summary><strong>Crash Report</strong></summary><pre><code>

</code></pre></details>

**Additional Information**

ThomasWilshaw:
Are you able to provide a stack trace?

aeroc7:
> Are you able to provide a stack trace?

Yes, updated. Just not sure how useful it is, as the stack trace actually seemed to differ from the several times I tried it, which is what led me to believe something wasn't getting shutdown in the proper order.

PR #1913, opened 2022-05-04
Refactor/simplify the threadpool implementation for render tickets

aeroc7:


itsmattkc:
Does this resolve #1899?

aeroc7:
> Does this resolve #1899?

No, that's a separate issue.

itsmattkc:
Understood. This looks good, my only concern is whether this might introduce race conditions related to RenderTickets. RenderTicket [emits a Qt signal](https://github.com/olive-editor/olive/blob/master/app/threading/threadticket.cpp#L127) when the job is finished, and when using QThreads, this signal is more or less guaranteed to be thread safe (if the sender and receiver are in different threads, Qt will queue the signal on the receiver's event loop so that the receiving function will run in the receiver object's thread, preventing race conditions). This was the purpose of the `ticket->moveToThread` calls. I'm wondering if without QThreads that will no longer be the case, and non-threadsafe receiver functions will now be run in other threads.

If this is the case, the easiest/simplest solution would be to add `Qt::QueuedConnection` to the signal connects, which would force the aforementioned event loop behavior. Otherwise if there's a non-Qt alternative, that could be worth exploring too.

aeroc7:
> Understood. This looks good, my only concern is whether this might introduce race conditions related to RenderTickets. RenderTicket [emits a Qt signal](https://github.com/olive-editor/olive/blob/master/app/threading/threadticket.cpp#L127) when the job is finished, and when using QThreads, this signal is more or less guaranteed to be thread safe (if the sender and receiver are in different threads, Qt will queue the signal on the receiver's event loop so that the receiving function will run in the receiver object's thread, preventing race conditions). This was the purpose of the `ticket->moveToThread` calls. I'm wondering if without QThreads that will no longer be the case, and non-threadsafe receiver functions will now be run in other threads.
> 
> If this is the case, the easiest/simplest solution would be to add `Qt::QueuedConnection` to the signal connects, which would force the aforementioned event loop behavior. Otherwise if there's a non-Qt alternative, that could be worth exploring too. I'm not 100% sure if this is actually occurring or not at this time though.

Hm, I think it's fine, but I'm no Qt expert. Here's what I found:

> [...signal emission is thread-safe.](https://doc.qt.io/qt-5/threads-qobject.html#accessing-qobject-subclasses-from-other-threads)

And if I'm not mistaken, the only `connect` that is made to the thread ticket slot [Finished](https://github.com/olive-editor/olive/blob/0e3d6ef37429c6ab9cd3f29b906bd581a7e5f92a/app/threading/threadticket.h#L114) is in [threadticketwatcher.cpp](https://github.com/olive-editor/olive/blob/0e3d6ef37429c6ab9cd3f29b906bd581a7e5f92a/app/threading/threadticketwatcher.cpp#L48). There is no specified connection type in this call, so the default is [Qt::AutoConnection](https://doc.qt.io/qt-5/qt.html#ConnectionType-enum), which, according to the documentation, automatically determines if it can be a direct call or queued call (for thread safety).

`Qt::AutoConnection`
> (Default) If the receiver [lives in](https://doc.qt.io/qt-5/qobject.html#thread-affinity) the thread that emits the signal, Qt::DirectConnection is used. Otherwise, Qt::QueuedConnection is used. The connection type is determined when the signal is emitted.

itsmattkc:
Yeah but I have a feeling that behavior relies on `QObject::thread()` which is set with `QObject::moveToThread`. See the notes on that same page:

> By default, a QObject lives in the thread in which it is created. An object's thread affinity can be queried using [thread](https://doc.qt.io/qt-5/qobject.html#thread)() and changed using [moveToThread](https://doc.qt.io/qt-5/qobject.html#moveToThread)().

> Note: If a QObject has no [thread](https://doc.qt.io/qt-5/qobject.html#thread) affinity (that is, if thread() returns zero), or if it lives in a thread that has no running event loop, then it cannot receive queued signals or posted events.

Is there a way to test which of these threads the receiver function is being called in? (in Qt it would be `QThread::currentThread` but I assume that wouldn't be applicable here)

aeroc7:
> Yeah but I have a feeling that behavior relies on `QObject::thread()` which is set with `QObject::moveToThread`. See the notes on that same page:
> 
> > By default, a QObject lives in the thread in which it is created. An object's thread affinity can be queried using [thread](https://doc.qt.io/qt-5/qobject.html#thread)() and changed using [moveToThread](https://doc.qt.io/qt-5/qobject.html#moveToThread)().
> 
> > Note: If a QObject has no [thread](https://doc.qt.io/qt-5/qobject.html#thread) affinity (that is, if thread() returns zero), or if it lives in a thread that has no running event loop, then it cannot receive queued signals or posted events.
> 
> Is there a way to test which of these threads the receiver function is being called in? (in Qt it would be `QThread::currentThread` but I assume that wouldn't be applicable here)

After looking through the Qt source code, here's what I found:

The [thread id implementation](https://github.com/qt/qtbase/blob/f087e63bf585f51d6c530dd58a57155835e6df2c/src/corelib/thread/qthread.h#L187) is just a wrapper around the OS-specific method of getting the thread id. They seem to use some custom assembler, but an equivalent call, if using pthreads, would be `pthread_self()` (which they actually seem to fall back to on platforms they can't use assembler for anyways).

So it's not Qt-specific. In fact, all the calling system is doing is checking if the thread ids are the same, and in any case they aren't, it's a queued call. Seen [here](https://github.com/qt/qtbase/blob/a3f7dd5260d1855e637d4a616a82c96b4e4cc77c/src/corelib/kernel/qmetaobject.cpp#L1520).

However, it might be worth it, for clarity sake, to add an explicit `Qt::QueuedConnection` to the `connect` call.

itsmattkc:
Ahh okay great, thanks for looking into that and confirming how that works. That was my only concern, so this should be good to merge

PR #1915, opened 2022-05-05
Use multiple build-jobs in CI builds to (hopefully) speed up the runs

aeroc7:
Noticed the CI builds were taking ~10+ minutes each. I understand the CI machines are only ~2 cores, but this should give some improvement.


aeroc7:
Oops, my mistake. It seems that Ninja is being used, which already builds in parallel by default (I'm used to make). Not sure if this will have any noticeable impact after all, as I believe CMake just passes this variable to the underlying build tool.

PR #1920, opened 2022-05-08
Implement RGB to B&W node

ThomasWilshaw:
Implements a RGB to B&W node using OCIO luma coefficients. ALso allows for overriding these with custom weights. There's a few questions in the code comments.

Specifically is there a good way to have coloured sliders be disabled and greyed out? I feel my solution is not ideal. Also I've tried to cache the luma coefficients but I can't seem to get it to work properly.

### DIFF
--- a/app/node/color/rgbtobw/CMakeLists.txt
+++ b/app/node/color/rgbtobw/CMakeLists.txt
@@ -14,13 +14,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-add_subdirectory(colormanager)
-add_subdirectory(displaytransform)
-add_subdirectory(ociobase)
-add_subdirectory(ociogradingtransformlinear)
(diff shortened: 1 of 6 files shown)


sobotka:
Not sure if you do this with this patch, but it may be wise to expose the weights and fill them with the default as determined. This would allow cases where folks may require alternative weighting.

ThomasWilshaw:
Yes there's a checkbox to use custom weights included.

ThomasWilshaw:
@itsmattkc quick question, the title of this node is `RGB to B&W` which displays fine in the properties viewer. However, in the drop downs (e.g. the add effect drop down) Qt uses the `&` as an emphasis and so it displays as "RGB to B<ins>W</ins>" with the `W` underlined (see image). If I use a double ampersand to correct this it displays incorrectly in the node header. What do you think the best approach is here? Just using one ampersand and living with the underline o rmaybe even just ditching the ampersand completely?

Other than that I think this is good to go.

image

PR #1974, opened 2022-07-17
Text animation

WileECoder:
The purpose of this PR is to provide a basic text animation capability, inspired by After Effects text animators but not so powerful at all.
It is possible to animate the position and rotation of each letter of the text generated by a Text node.
[I have made a video](https://youtu.be/Azj_nMPygEw) to show how it is supposed to be used.


itsmattkc:
Hey this is really cool, and I really appreciate the work you've put in to make this work. I have some concerns about the implementation though.

Firstly Olive's nodes are designed so they can push multiple values and later nodes can choose which one they want, so an "Output HTML" option should be unnecessary.

Also what's the purpose of having separate text animation and text animation render nodes? Is the idea to allow for future modularity?

WileECoder:
> Firstly Olive's nodes are designed so they can push multiple values and later nodes can choose which one they want,

I was wishing for this. So I can get the text and also the base position, and this saves an input in the animation node. However it's not quite clear to me how the receiver node selects the value, as the selection key is only the (one) string that defines the input.
Can you point an example of a node that does this? 

> Also what's the purpose of having separate text animation and text animation render nodes? 

The reason why I chose to separate an animation descriptor to the renderer is mainly to allow composition of several animators for the same text.
The alternative to this may be an array input in the renderer, but I think this would be harder to manage and would make the number of inputs explode.
With a separate node, it is also possible to exclude one animator with the "enable" checkbox of the single animation node; it is also a bit simpler to save or share an animation preset, that someone else can add (or compose) to a "render" node that is already in the project.
If you prefer a different implementation, let me know.

WileECoder:
OK, I've figured it out. Now the Text node is unchanged and the base position of the text is taken from the Text node.

itsmattkc:
Hi, sorry it's taken a while to get back to this.

I'm still uncertain about the implementation. I found that the "text animation" node parameters weren't very clear or intuitive, and the "text animation render" node still feels superfluous (they may as well just plug into the regular "text" node).

Even then, I'm not sure if plugging text animators in and passing XML between the nodes is the best approach, though without knowing exactly the needs and goals of this system, I couldn't say what a better one is.

WileECoder:
>I'm not sure if plugging text animators in and passing XML between the nodes is the best approach, though without knowing exactly the needs and goals of this system, I couldn't say what a better one is.

The goal is to animate a text, or part of a text, in a manner that is modular and reusable, i.e:
 - several animations can be applied to the same text (or to different parts of a text)
 - an animation, (or a set of animations), can be used for other text nodes, maybe in different projects, that is why the animation is not embedded in the text node.

[I have made this video](https://www.youtube.com/watch?v=aq0BfN-zkh8) as a showcase of what can be achieved.

The reason why XML is used is that the input of a node must fall in one of the types defined in `NodeValue::Type`. There is no "binary data" type, so I thought that `kText` was the only type able to represent almost anything. Any text based format will do, but with XML we have a parser and formatter for free.
The alternative may be to add an entry to `NodeValue::Type`: this could be a generic `kBinaryData`, serialized as Base64 in the project file, or something specific for the text animation. Let me know if you think this implementation is better.

> I found that the "text animation" node parameters weren't very clear or intuitive

About this, maybe better names for inputs might be used (any suggestion is appreciated here). An alternative might be to have tool-tips for each input. In any case, if the PR is accepted, I can make detailed tutorials (video and written) and provide preset nodes to copy from.
A little trial and error is still needed.

> and the "text animation render" node still feels superfluous (they may as well just plug into the regular "text" node).

OK, that can be done. My idea was not to complicate the text node with a feature that one might not need, but if you think that it's better to add an "animator" input to the current text node, I will.

TheTechRobo:
Would it be possible to animate the font size? (Haven't looked closely at this PR, so sorry if this isn't possible or already done.)

WileECoder:
> Would it be possible to animate the font size? 

With the current implementation, you can modify the horizontal and the vertical stretch: this has the same effect of changing font size:

https://user-images.githubusercontent.com/109035347/195440566-f7cbd75c-ad84-4205-b58a-d21c510e35b9.mp4

WileECoder:
In this version the TextAnimationRendernode has been removed and the Text node has an input "animators".
If such input is not plugged, the node does exactly what it used to. Only if one (or more) animators are plugged, a different render function is used.
Even if one or more animators are plugged, when the user double clicks to edit text, the usual code is called.

About the inputs that are not intuitive, I have added a tool tip that shows when the user hovers the input: in the tool tip I have tried to explain the meaning of the input as best as I could.

tooltip

[I have also made a wiki page](https://github.com/WileECoder/olive/wiki/TextAnimation) to explain in detail the usage of each input in detail. This may also help to suggest better names ...

ThomasWilshaw:

WileECoder:
I still think this is useful and I still have time to work on this (though it's ready for review). However I appreciate any feedback on how interesting this is.

WileECoder:
As you can read on Patreon page, the lead developer of Olive project has
paused his activity in search for investors interested in the project.
Merging or reviewing a PR like this is not currently a priority at all.

mcDandy:
There Is no Patreon on olive main page. Adding it might help.

Dne pá 3. 11. 2023 21:07 uživatel WileECoder ***@***.***>
napsal:

> As you can read on Patreon page, the lead developer of Olive project has
> paused his activity in search for investors interested in the project.
> Merging or reviewing a PR like this is not currently a priority at all.
>

itsmattkc:
> There Is no Patreon on olive main page. Adding it might help.

There used to be. But as I say in the last post, I don't think the kind of money we need can be raised through crowdfunding. Getting what we have now to something that could support not just one full time developer but ideally a team is an exponential difference. The current site is designed to be more focused on showcasing Olive's plans and innovations.

PR #1987, opened 2022-07-27
nclc tags

ThomasWilshaw:
This is a quick test of adding nclc tags and a Rec.709 OETF for rendering out of Olive. @sobotka and @itsmattkc would you be able to have a look at this and test it on MacOS as I can't seem to find a player on Windows that actually respects these tags properly


itsmattkc:
I've made the following adjustments:

- Removed MOV restriction, since QuickTime at least can pick up these tags in other containers like MP4 too.
- Made the `sRGB` check case insensitive for configs that may write it `srgb`
- Changed `AVCodecContext::colorspace` attribute to `AVCOL_SPC_RGB` for sRGB, which according to FFmpeg documentation, is the correct one for IEC 61966-2-1.

sRGB videos now appear to display correctly in QuickTime (and FFplay) with no color shift (on both limited and full YUV), however the Rec.709 export still appears to exhibit some.

Here's a comparison of the two tagged exports in QuickTime, sRGB (correct) on the left and Rec.709 (incorrect) on the right:
Comparison of two different exports

Presumably this is either an issue with the tagging still or an issue with LUT?

ThomasWilshaw:
Interesting that it seems to work with those tags, I was getting a black screen in QuickTime when I used my sRGB tags even though it was the same as Baselight tags sRGB. It would be good to try say rendering and tagging correctly with P2 D65 and the sRGB EOTF.

I suspect the issue is with either the output lut or just how we're handling applying luts on rendering. For example if you set the outptu CS to Apple DCI-P3 D65 I geta  big colour shift but setting it to Apple P3 sRGB EOTF it looks the same as the standard sRGB EOTF even though there should be a gamut map going on

sobotka:
More folks should be aware of this video, so posting it here for some good details. https://vimeo.com/349868875

ThomasWilshaw:
As far as I can tell (by running some values through the repsective luts) both the sRGB OETF and Rec.709 EOTF are working correctly and giving the right values. The question is then whether the tagging is correct. What we need to remember is that a correctly managed Rec.709 image and a correctly managed sRGB image will not match code value for code value as they are inteded to be viewed in different surrounds. They shold match visually when the viewer has adpated to their surround. 

Going off of Apples [documentation](https://developer.apple.com/library/archive/technotes/tn2162/_index.html#//apple_ref/doc/uid/DTS40013070-CH1-TNTAG10) I think the correct tagging for Rec.709 should be:
`1-1-1`

And as far as I can tell from other ISO spec [papers](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwjQ2cGWiKH5AhUNhlwKHX7KBY8QFnoECBIQAQ&url=https%3A%2F%2Fwww.itu.int%2Frec%2Fdologin_pub.asp%3Flang%3Df%26id%3DT-REC-H.273-202107-I!!PDF-E%26type%3Ditems&usg=AOvVaw1A2fWTEqSSholk39bR_j-x) the sRGB should be:
`1-13-1`

I'm really not sure where ffmpeg got their information from, I've not seen that anywhere else and `0` for the matrix is usually reservered or a identity matrix.

I might try and generate some test clips with known values for you test in Quicktime and colour pick from.

sobotka:
Invaluable information here for future astronauts. To those astronauts who may arrive here in the distant future:

Please appreciate that this terrain is muddied if we fail to contextualize the usage cases!

There are two extremely important and somewhat divergent designs that future astronaut archaeologists must discover in order to understand the nuance and complexity here.

The first, is to appreciate our historical context of electrical engineering and the need to replicate general “appearance”. This is firmly anchored in the all-too-rarely discussed viewing context. Specifically, the *intention* of the BT.709 encoding standard never stipulated a viewing output context, nor EOTF. This discrepancy was later addressed due to the hard work of some individuals and resulted in BT.1886, which broadly specifies a 2.4 pure power function as the EOTF and a dark viewing condition.

The discrepancy between the *encoding* and the *decoding* is intentional; the tristimulus ratios are subtly distorted. In distorting, there is a subtle variation in “contrast” as well as sensation of “chroma”. This is *by design*. In an attempt to provide an appearance match of intention, the tristimulus values *must* be distorted with respect to their origin. This distortion was historically known as “system gamma”; the total overall estimate deviation of source encoding to destination after full encoding to decoding.

We can see echoes of this design by examining the sRGB reference, of which a reference to the BT.709 document is made, highlighting how the lack of an implied EOTF is a point of problem. To this end, the sRGB document contains both an assumed encoding OETF that is composed of a similar two part encoding, and a reference EOTF composed of a single power function of 2.2. So why the discrepancies? The sRGB standard stipulates a rather loose definition of “dim” viewing conditions, as compared with BT.1886’s “dark”.

Informed folks will also cite another case of digital cinema’s implied 2.6 power function for an EOTF. While I can only offer up non-cited anecdotal evidence due to anonymity pledges, it was suggested that the 2.6 result for cinemas was the value that yielded the best appearance match for a dark theatre situation where the maximal achromatic tristimulus output achieves the standard 48 nits.

The second part of this is that when authors are authoring content in a Digital Content Creation application, the needs of their internal representation of the information will be different for performing algorithms and manipulations, when compared to the above details of historical presentation.

That is, how we should decode something for working on, versus how we should encode something for viewing / consumption, rest entirely on different surfaces of complexity. It would be a foolish individual who would suggest that we should treat all cases with the same approach!

There are dark and mysterious appearance phenomenon at work, here too. Specifically, with respect to the expansion and contraction of “contrast”, it is worth researching contrast during adaptation. The Hunt Effect is also worth researching for the influence of general shifts in chroma at varying adaption levels as well as tristimulus amplitudes.

Finally, if folks are interested in this subject, part of the historical framework is wonderfully outlined by [Filmlight’s Daniele Siragusano in the “sRGB… We Need to Talk” presentation](https://m.youtube.com/watch?v=NzhUzeNUBuM)

ThomasWilshaw:
For those interested

system_gamma

PR #1995, opened 2022-08-03
Added option for user to disable transparent dialogs

p9malino26:
Apologies for re-creating this but because I am still learning how to use GitHub, I've had to re-order my branches, forcing the previous pull request #1984 to be deleted and me to duplicate this pull request.
~~~
So I ran Olive in I3WM and had a problem when editing title text in video viewer: the text I typed did not show up at all in the editor, but it did show up on the video after I quit the Viewer Text Editor. This was because I3WM has issues with transparent windows. I did not get this issue when using KDE Plasma.

Therefore, I have added an option into Preferences which allows the user to turn off/on transparent dialogs, which is enabled by default but can be disabled in case of any problems.

Can provide screenshots on request.

### DIFF
--- a/app/config/config.cpp
+++ b/app/config/config.cpp
@@ -159,6 +159,8 @@ void Config::SetDefaults()
   SetEntryInternal(QStringLiteral("OfflinePixelFormat"), NodeValue::kInt, VideoParams::kFormatFloat16);
 
   SetEntryInternal(QStringLiteral("MarkerColor"), NodeValue::kInt, ColorCoding::kLime);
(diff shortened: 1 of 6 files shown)


itsmattkc:
You're not going to like this, but I think this is a fairly niche option for a very specific use case and doesn't really need a whole dedicated UI option. At best, it could go in the "Behavior" section, where a lot of minor options go.

p9malino26:
So I now added the option into the Visual section under General:
image

itsmattkc:
Seeing this (and various other issues) I wondered if it was possible to reimplement Olive's text editor without relying on window manager transparency. I've managed to do so in the latest commit which should negate the need for this option.

PR #1996, opened 2022-08-03
Fix viewer text editor going outside screen edge

p9malino26:
I have made the viewer text editor be clamped to edges of the screen so that it does not go out of view.
Video demonstration attached (text not visible due to I3WM bug which I've fixed in another pull request that is yet unmerged).
[Video](https://user-images.githubusercontent.com/29703546/182684637-5ebfd74f-a716-4ccb-b3e0-37f140eb071a.webm)


itsmattkc:
Thanks for the PR, but with the recent changes to the text editor, this is no longer necessary

PR #2030, opened 2022-09-16
polygon: Implement polygon gizmo selection and dragging

patryk-m-malinowski:
Changes are:
- Ability to select multiple points of polygon by holding down CTRL and then dragging all the points at once
- Bezier lines only show for selected gizmo when only one gizmo is selected
- Double clicking deselects all points

https://user-images.githubusercontent.com/29703546/190712378-4bfae786-f9f6-4e3c-91ca-86b3ef8c00a9.mov

This also fixes the bug where dragging point gizmos does not update the GUI immediately.

Issue #1907, opened 2022-04-29
[UI] Improve polygon gizmos


itsmattkc:
**Summary**
Polygon are a little more cumbersome to edit than they should be with the current gizmos. This is a checklist of what I'd like to see:

- [ ] Allow polygon points to be selectable so that several can be moved at once
- [ ] Set reasonable defaults for bezier points if the user wants to use them
- [ ] Only show bezier points for selected points - tying into step 1, and allowing for less UI clutter/confusion
- [ ] After step 3, bezier points can be made bigger and more clickable because they're less likely to conflict/be confused with other points

unfa:
I have some more issues with the polygon / mask editor gizmos:

1. The editing lacks visual feedback (hilighting nodes that are close enough to move)
2. Editing requires pixel-perfect mouse inputs (no snapping distance to register a mouse click)
3. There's no way to decide if you want to grab a bezier handle or the point. They start out all stacked on top of one another and it seems random when you grab one - a modifier key could solve this.
4. Polygon editing gizmo doesn't update until you move the playhead (the mask one does though ,strangely enough)
5. Animating points should insert a keyframe for all points when triggered on the top level, and then auto-insert points for all nodes in sync - otherwise the curve points would be drifting between shapes and htat'd be unexpected.
let the users tweak the low-level per-point keyframes if they want, but make it easy to insert a keyframe set fro the entire curve with one click, and then auto-keyframe all points when one of them is changed (maybe a checkbox for that would be good?)

p9malino26:
I'm working on it.

![image](https://user-images.githubusercontent.com/29703546/188430741-9d7acd74-df9a-44b9-935f-029ea68007a3.png)
![image](https://user-images.githubusercontent.com/29703546/188430929-f50c70a3-cbd4-4810-9b17-e6c40b19d4e2.png)

PR #2034, opened 2022-09-21
Another polygon improvement from Issue #1907

WileECoder:
This PR addresses issue #1907 and improves the behaviour of Bezier polygon (and the Mask node).
Both issues from @itsmattkc and @unfa are addressed; in particular:

- a set of points can be selected, and more points can be added or removed to selection
- selection can be edited by clicking single points or with a lasso selection
- clicking inside the polygon moves the whole polygon, consistently with what happens for Text and Shape
- only handles for selected points are shown
- there is a visual feedback when a gizmo is hovered and ...
- ... points are bigger and easier to select
- if a Bezier point and its control point are in the same position, dragging always affects the main point.
   In order to grab a control point, ALT modifier can be used. If control point is isolated, ALT modifier is not needed.

I have made a video to show how this works.

NOTE: This PR duplicates the one from @p9malino26 (#2030). I did not realize he was working on the same issue.
 Feel free to use the best implementation or take something good from both ...


ThomasWilshaw:

WileECoder:
While I still think this is a useful feature, chances are that a refactoring will generate too many conflicts. Fell free to close this.

PR #2047, opened 2022-10-01
Add community unofficial matrix

uranuspucksaxophone:
Closes #2000


itsmattkc:
To be honest, I don't really think it's a good idea to promote a community that we have no involvement with.

PR #2077, opened 2022-10-25
ffmpegdecoder: Catch bad stream duration estimates

### LINKED ISSUE #2076: [EDIT] Olive/FFmpeg fails to correctly detect MXF duration
**Commit Hash** 
[3d90dc4](https://github.com/olive-editor/olive/commit/3d90dc4ecd61ee67d3e52024695b99c0898256cd)

**Platform** 
Windows 10

**Summary**
Olive fails to properly detect the duration of Arri Alexa 35 sample clips. For example [this](https://webgate.io/directlink/bdb5241a98e21e39/folder/680418) clip is 10 seconds and 13 frames long but Olive interprets it as being about 34 minutes. 

As far as I can tell this is actually a problem with ffmpeg incorrectly decoding the file but I'm wondering if we can find some way to confirm the duration of a clip. The fall back method that is already in the code does produce the correct results but is not used in this case as Olive thinks it has correctly decoded the file. Always calling the fallback method would be bad as it can be very slow on long clips so it would be good to find some way of only callin this when needed. 

Iterestingly MediaInfo reports the duration correctly so it is possible.

ThomasWilshaw:
For some formats, particularly mxf, FFmpeg calculates the duration of the stream incorrectly. Here we try to catch that and force Olive to use it's much slower, but correct fallback method.

It would be nice if it were possible to inform the user that a (potentially much) slower method was being used.

Fixes #2076

### DIFF
--- a/app/codec/ffmpeg/ffmpegdecoder.cpp
+++ b/app/codec/ffmpeg/ffmpegdecoder.cpp
@@ -350,6 +350,13 @@ FootageDescription FFmpegDecoder::Probe(const QString &filename, CancelAtom *can
 
     int64_t footage_duration = fmt_ctx->duration;
 
+    bool bad_duration = false;
+
+    if (fmt_ctx->duration_estimation_method == AVFMT_DURATION_FROM_BITRATE) {
+      bad_duration = true;
+      qWarning() << "Potentially bad duration estimation, using fallback. This could be slow.";
+    }
+
     // Dump it into the Footage object
     for (unsigned int i=0;i<fmt_ctx->nb_streams;i++) {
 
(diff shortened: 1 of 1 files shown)


app/codec/ffmpeg/ffmpegdecoder.cpp
```diff
@@ -350,6 +350,15 @@ FootageDescription FFmpegDecoder::Probe(const QString &filename, CancelAtom *can
 
     int64_t footage_duration = fmt_ctx->duration;
 
+    bool bad_duration = false;
+
+    // Catch when ffmpeg uses its inaccurate method of duration estimation so we can 
+    // use manual calculation
+    if (fmt_ctx->duration_estimation_method == AVFMT_DURATION_FROM_BITRATE) {
```
itsmattkc, line 357: This might be a little pedantic, but I would prefer a more descriptive variable name than `bad_duration`. Maybe something like `duration_guessed_from_bitrate`. You could also make this tidier by simply setting it from the initialization (i.e. `bool duration_guessed_from_bitrate = (fmt_ctx->duration_estimation_method == AVFMT_DURATION_FROM_BITRATE);`)

ThomasWilshaw: Thanks for the feedback, much appreciated. I've updated as you suggested

PR #2082, opened 2022-11-01
NSIS: Use 64-bit version of ShellExecAsUser.dll

### LINKED ISSUE #2075: Strange error message on first start
**Commit Hash** 
3d90dc4e

**Platform** 
Windows 10

**Summary**
On first startup (I started Olive from the installer), I got this strange error message. I figured I should mention it, since I'm not sure if it's just a fluke, or a problem with a module/setup.

image

Simran-B:
The `Amd64 Unicode` plugin download from https://nsis.sourceforge.io/ShellExecAsUser_plug-in contains an ANSI version, a 32-bit Unicode version, and a 64-bit Unicode version. The CI currently uses the 32-bit version.

My hope is that using the 64-bit version resolves #2075, but the problem could also be deep inside its source code or how it was built (the readme mentions VS 2010...)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,8 +259,8 @@ jobs:
           cp $(cygpath $GITHUB_WORKSPACE)/app/packaging/windows/nsis/* .
           cp $(cygpath $GITHUB_WORKSPACE)/LICENSE .
           $DOWNLOAD_TOOL https://nsis.sourceforge.io/mediawiki/images/6/68/ShellExecAsUser_amd64-Unicode.7z
-          7z e ShellExecAsUser_amd64-Unicode.7z Plugins/x86-unicode/ShellExecAsUser.dll
-          makensis -V4 -DX64 "-XOutFile $PKGNAME.exe" "-X!AddPluginDir /x86-unicode $(pwd -W)" olive.nsi
+          7z e ShellExecAsUser_amd64-Unicode.7z Plugins/amd64-unicode/ShellExecAsUser.dll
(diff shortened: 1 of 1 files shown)


Simran-B:
Looks like the 64-bit DLL does not get picked up:

```
7    7z e ShellExecAsUser_amd64-Unicode.7z Plugins/amd64-unicode/ShellExecAsUser.dll
8    makensis -V4 -DX64 "-XOutFile $PKGNAME.exe" "-X!AddPluginDir /amd64-unicode $(pwd -W)" olive.nsi
...
92   Processing default plugins: "C:\Program Files (x86)\NSIS\Plugins\x86-unicode\*.dll"
...
159  !addplugindir: "D:/a/olive/build" (amd64-unicode)
...
180  Processing default plugins: "C:\Program Files (x86)\NSIS\Plugins\x86-unicode\*.dll"
...
381  Plugin directories:
382    C:\Program Files (x86)\NSIS\Plugins\x86-unicode
```

The official NSIS is apparently available in 32-bit only, and I confirmed that the plugins it ships with are in `x86-ansi` and `x86-unicode`. I'm not sure what the 64-bit plugin even is for, maybe custom 64-bit builds of NSIS?

itsmattkc:
Do you encounter this behavior consistently?

Does it only occur with the ShellExecAsUser plugin? If you revert to ExecShell does the behavior go away?

That could help us narrow down or rule out a potential cause.

ThomasWilshaw:

Simran-B:
This was only a test, and I realized that NSIS isn't a 64-bit (unless you manually compile it that way). I couldn't reproduce the problem locally so far.

Issue #1844, opened 2022-01-24
[RENDER] Pixel aspect ratio issues


itsmattkc:
**Summary**
Pixel aspect ratio handling appears to have regressed since it was last finished. Particularly in regards to the following nodes:

- [ ] Transform
- [ ] Shape
- [ ] Text
- [ ] Polygon

Check interlacing too.

PR #2084, opened 2022-11-02
viewertexteditor: Fix incorrect text rendering when pixel aspect ratio not 1

patryk-m-malinowski:
Addresses issue #1844, although it solves the problem with text only and not polygon, shape, and transform

### DIFF
--- a/app/node/generator/text/textv3.cpp
+++ b/app/node/generator/text/textv3.cpp
@@ -150,12 +150,12 @@ void TextGeneratorV3::GenerateFrame(FramePtr frame, const GenerateJob& job) cons
 
   // Draw rich text onto image
   QPainter p(&img);
-  const double par = frame->video_params().pixel_aspect_ratio().toDouble();
-  p.scale(1.0 / frame->video_params().divider() / par, 1.0 / frame->video_params().divider());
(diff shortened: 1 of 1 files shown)


ThomasWilshaw:

p9malino26:
It's up to you whether to merge this or not. I think this may be useful. If not then feel free to close this.

p9malino26:
It's up to you. I think this may be useful, but feel free to close this if not.

app/node/generator/text/textv3.cpp
```diff
@@ -150,11 +150,12 @@ void TextGeneratorV3::GenerateFrame(FramePtr frame, const GenerateJob& job) cons
 
   // Draw rich text onto image
   QPainter p(&img);
-  p.scale(1.0 / frame->video_params().divider(), 1.0 / frame->video_params().divider());
+  const double par = frame->video_params().pixel_aspect_ratio().toDouble();
```
ThomasWilshaw, line 153: Long explicite names are prefered so something like `pixel_aspect_ratio` would be better

patryk-m-malinowski:
Closed due to inactivity.

PR #2099, opened 2022-11-13
Alexa35

ThomasWilshaw:
Adds both the NTSC and PAL Rec spaces and the Alexa 35 input space. The Alexa 35 to linear transform comes from the ACES configs.

@sobotka, could you cast an eye over this please?


sobotka:
I would desperately avoid the ACES values and go directly from the Alexa35 white paper, toward a CIE defined colourimetry, ideally anchored in Illuminant E. Then base all transforms to be dynamically outward from there.

ThomasWilshaw:
I've only used the LogC4 to linear transform from the ACES config. I had generated a LUT from the white paper but this seemed cleaner

sobotka:
Not sure where color.support is getting their matrices from. The Horse’s Mouth is available from the Sean Cooper (ARRI) transform matrix using CAT02 as the reprojection to ACES fuckoD60, which can be undone.

If memory serves, the entire ARRI chain wisely rallies around a clear D65 definition, and betting on the CAT02 transform being everywhere, it can be reasonably take to the Standard Observer IE basis for future work outward.

It should also be noted somewhere in the code that the 3x3 approach is superseded by ARRI’s black box code in their Reveal chain.

https://github.com/ampas/aces-dev/blob/dev/transforms/ctl/idt/vendorSupplied/arri/alexa/v4/IDT.ARRI.ARRI-LogC4.ctl#L41-L43

ThomasWilshaw:
Yeah it's listed here http://color.support/colorspacecalculator.html as AWG4 and it uses the same white point as REC.709 so I would have thought we dnlt need any white point reprojection? 

I didn't mean to confuse anything by mentioning ACES, I just meant I pulled [this](https://github.com/olive-editor/olive/pull/2099/files#diff-6a6583b61d4a30d1c40b35c51e1a351f6cf4096654318bc49d7bf75a0c752b2cR280) line out of the config to avoid having another 1D lut

sobotka:
Usually ARRI releases white papers on their colourimetry, including the log transforms. Sean’s source code is as reliable as it gets.

I am not sure if the V4 colourimetric papers have made it down the chain yet?

It is quite correct that no reprojection is needed. The reason I suggested that the baseline might be worth keeping around is that the camera quantal catches are fit to XYZ colourimetry, sometimes using an interstitial model.

The “error” and nature of the fit is entirely anchored in that interstitial model, which is often times not specified, and vary model to model. There is often times calculation bias in the underlying math due to the predicate of illuminant, and how the Standard Observer IE model is constructed with the Illuminant achromatic axis not being the centroid.

ThomasWilshaw:
Yes the white paper is available [here](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwi4lN-f6Kv7AhUVSUEAHd_6Al0QFnoECBMQAQ&url=https%3A%2F%2Fwww.arri.com%2Fresource%2Fblob%2F278790%2Fbea879ac0d041a925bed27a096ab3ec2%2F2022-05-arri-logc4-specification-data.pdf&usg=AOvVaw304u76ZpQ52YebU6KGhiob) and is what I originally used to generate the LUT.

The paper does give the AWG4 to XYZ matrix but I donlt really understand what a CAT02 is other than a way of chaning the white point of a colour space

sobotka:
Nice.

> The paper does give the AWG4 to XYZ matrix but I donlt really understand what a CAT02 is other than a way of chaning the white point of a colour space

The entire concept of colourimetry is based around opponency, and a theist belief in Grassman additivity extended to the infinite amplitude case, which is of course goofy as shit.

The CAT02 is a simple matrix, much like Bradford, that massages the Standard Observer “primaries” such that the results are closer to yet another arbitrary error metric. In this case, the matrix is applied to uniform tristimulus relative to the Standard Observer, and reprojects it such that the results are similar to “complete adaptation”.

The “proper” Bradford of has a nonuniform adjustment to the amplitudes, which we would expect given that the lowest order of our perceptual chain transforms the radiometric signal amplitudes to something specifically oriented around our visual processing.

TL;DR: Think or modified Bradford and CAT02 as nothing more than a calculated matrix hand wavy magic to reproject the signal in an attempt to achieve something that resembles “chromatic adapatation” under our colourimetric models.

ThomasWilshaw:
Thanks for that, so are you saying it would be better to do a AWG4 -> XYZ -> Rec.709 stack in the config to avoid anything that might be going on under the hood in a combined matrix?

sobotka:
With OCIO V2 those would be concatenated anyways.

Ultimately it is about lowering the labour cost. I believe rather strongly that showcasing each and every transform in your code, with comments, aids readability, which in turn can aid contributions. Contributions here, including error checking or “improvements” are worth their weight in gold, even if they are pure auditing via GitHub or such.

As for labour, I also believe that keeping the matrices clean, and uniformly aligned with CIE colourimetry can be beneficial for lowering the cost of labour. It might also help in forward looking changes.

PR #2160, opened 2023-02-10
Compiling doesn't work because 'core' library is needed

balkNfalkN:
Currently the only way to build the codebase is to first build and install the [core](https://github.com/olive-editor/core) repo.
While this isn't hard to do, the bigger issue here is that it's unclear which revision of [core](https://github.com/olive-editor/core) would be safely compatible with which revision of [olive](https://github.com/olive-editor/olive). Git submodules could help here.

**This PR is just a reference to how you might produce an AppImage currently**.

Another thing is I found you need to compile with `clang++` otherwise errors emitted due to `-Wpedantic`

Just type:

`./build_appimage.sh`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "core"]
+	path = core
+	url = https://github.com/olive-editor/core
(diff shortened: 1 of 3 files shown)


itsmattkc:
Thanks for the concept PR. The core library has been indeed added as a submodule and is built automatically by CMake. Closing this since it has now been implemented.

Issue #1630, opened 2021-05-12
[UI] Wayland Meta Thread


itsmattkc:
We don't currently have the means to properly support Wayland and there appear to be several issues when using Olive natively on Wayland rather than X11 or XWayland. To prevent clogging up the issue tracker, I'm making this thread as a general list of known issues on Wayland that people can post in and add to instead. If you're an expert with Wayland, feel free to contribute pull requests to solve any of these issues.

- Viewer/render issues. Textures will show other parts of the desktop rather than the video being rendered. Possibly an OpenGL context issue. (#1507)
- Action search does not appear to respond to keyboard input. (#1629)

If you find other Wayland-exclusive issues, comment here and they will be added to this main post.

sobotka:
Start a “Discussions” thread and merge there?

mara004:
@elsandosgrande Could you please check whether you are experiencing #1629 on Gnome Wayland, too?

mara004:
Maybe we could clarify point 1 a little: It's not that the video to be viewed would be completely invisible, only the background flickers and is transparent, but the video itself is rendered correctly. Maybe that was not clear enough from my screenshots and description.

mara004:
I just tested ae3393f, but it didn't fix #1507.

mara004:
The following is not a real issue since it does not impact the user, but when using Olive as Wayland client, I frequently get this warning in console:
```
[WARNING] setGrabPopup called with a parent, QtWaylandClient::QWaylandXdgSurface(0x55a5d8f499a0) which does not match the current topmost grabbing popup, QtWaylandClient::QWaylandXdgSurface(0x55a5debc3470) According to the xdg-shell protocol, this is not allowed. The wayland QPA plugin is currently handling it by setting the parent to the topmost grabbing popup. Note, however, that this may cause positioning errors and popups closing unxpectedly because xdg-shell mandate that child popups close before parents ((null):0)
```
For instance, this message shows up when triggering the pop-up explaining that you can't export an empty sequence. Maybe (probably) it's with all pop-ups, I haven't tried.

HarshilPatel007:
@itsmattkc , I've got crash on startup.
here is the DEBUG output.
```
➜  Desktop ./Olive-2794730a-Linux-x86_64.AppImage
[INFO] Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway. ((null):0)
[...]

PR #2226, opened 2023-05-19
Fix Wayland Transparency

Quackdoc:
Partially addresses #1630

it turns out that the alpha we set is actually passed onto the compositor, so by setting 0.0 alpha, we were inadvertently telling the compositor to make the window transparent.

### DIFF
--- a/app/render/renderer.h
+++ b/app/render/renderer.h
@@ -84,7 +84,7 @@ public:
 
   virtual void PostInit() = 0;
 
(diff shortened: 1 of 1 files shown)


Quackdoc:
turns out this might be a bit more tricky

itsmattkc:
I recently started playing around with Wayland and resolved a number of issues including this one. Thanks for finding the source of the transparency issue, but this should be resolved now.

PR #2371, opened 2025-02-15
linux/appdata: Fix description translations, add launchable tag, fix screenshot

AsciiWolf:
The _p_ tags inside description should be translated, not the _description_ tag itself.

(This fixes the default description being incorrectly in French instead of English [on Debian](https://appstream.debian.org/sid/main/metainfo/olive-editor.html).)

Also add a _launchable_ tag required by newer AppStream standard and some validators/parsers and fix the screenshot url.

Please see: https://github.com/flathub/org.olivevideoeditor.Olive/blob/master/patches/olive-metainfo.patch

/cc @Simran-B


AsciiWolf:
@Simran-B Just a small ping. :-) It would be great if this could get merged before the new Olive release.

AsciiWolf:
I was informed that the "olive" Git repository is abandoned. Closing.

